Issue #84, opened 2014-05-20
Configurable location of directories


krlmlr:
To allow changing the layout later.

PR #85, opened 2014-06-04
Separate code from data and other stuff

### LINKED ISSUE #89: Distinguish between default configuration for new projects and substitutes for missing configuration items
Missing values might correspond to "legacy" behavior that we better turn off for new projects but still want to support for backward-compatibility reasons.

### LINKED ISSUE #91: Auto-update inst/defaults/global.dcf when bumping package version
Need to figure out a neat way. Perhaps `make bump-x` where x is the position in the version to increment, with the help of muxspace/crant.

krlmlr:
Rationale:
- To be able to exclude "other stuff" when searching in all files for code
  - Exclusion by file type is not enough if there are many non-R files -- they are looked at anyway (at least by RStudio), and this slows down the search
- I find myself constantly switching between `lib` and `src`, sometimes `munge`. As it is now, there is a lot of "other stuff" between these precious folders. Having them closer together would simplify things.

Candidates: `lib`, `munge`, `src`. Did I miss anything?

Requires ~~#84~~ #90.
- [x] Internal: Store files in tar archives also verbatim, to track changes
- [x] Script to create TAR archives (and also closes #91)
- [x] Change structure internally
- [x] Migration code (including version bump)
- [x] Tests for migration code
- [x] Closes #89 
- [ ] Double-check website, e.g. there is a screenshot that shows the old folder structure


krlmlr:
I received no feedback to this so far, and I wonder if this is the best thing to do. People seem fine with the directory structure as it is. I propose to keep the current directory structure and only extract otherwise useful changes from this and the other pull requests.

connectedblue:
I propose that this PR be closed.  Any ideas contained within would need to be re-raised in a new PR - I'm not sure this could be safely resubmitted with all the change made since it was originally produced

PR #140, opened 2015-08-06
Handle multiple sheet variable.names out of xls(x).reader()

### LINKED ISSUE #137: Error using data.table() with xls sources
Here is a [MRE](http://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example):

``` R
## Create a test project to reveal the issue
library(ProjectTemplate)
create.project('testxls')
setwd('testxls')

## Write a dummy xls data file
require(xlsx)
write.xlsx(data.frame(x=1), file = 'data/dummy.xls')

## Load project
load.project(override.config = list(data_tables = TRUE))
```

... which produces the following output:

```
Loading project configuration
Autoloading helper functions
 Running helper script: helpers.R
Autoloading cache
Autoloading data
 Loading data set: dummy
Loading required namespace: gdata
Loading required package: data.table
data.table 1.9.4  For help type: ?data.table
*** NB: by=.EACHI is now explicit. See README to restore previous behaviour.
Converting data.frames to data.tables
Error in get(data.set, envir = .TargetEnv) : object 'dummy' not found
```

famuvie:
Accounts for these specific readers within .load.data()

Fixes #137

### DIFF
--- a/R/load.project.R
+++ b/R/load.project.R
@@ -211,6 +211,15 @@ load.project <- function(override.config = NULL)
                      filename,
                      variable.name))
 
+        ## Most times, variable.name will correspond to one loaded object
+        ## but in the case of xls(x) readers, they always return
+        ## variable.name.Sheet1 ... variable.name.SheetN
+        if (grepl('xls', extension)) {
+          variable.name <- grep(paste(variable.name, 'Sheet', sep ='.'),
+                                ls(.TargetEnv),
+                                value = TRUE)
+        }
+
         data.files.loaded <- c(data.files.loaded, variable.name)
 
         break()


R/load.project.R
```diff
@@ -211,6 +211,15 @@ load.project <- function(override.config = NULL)
                      filename,
                      variable.name))
 
+        ## Most times, variable.name will correspond to one loaded object
+        ## but in the case of xls(x) readers, they always return
+        ## variable.name.Sheet1 ... variable.name.SheetN
+        if (grepl('xls', extension)) {
+          variable.name <- grep(paste(variable.name, 'Sheet', sep ='.'),
+                                ls(.TargetEnv),
+                                value = TRUE)
```
krlmlr: This looks brittle to me. Can the XLS reader (and also other readers that create multiple variables) return a list of variable names actually created?

I spent some time trying to create the LoadMyData package from ideas in ProjectTemplate. It's certainly possible to overengineer a solution for the data loading problem. I've come to the conclusion that the concept of automatically loading data only works for "standard" formats such as R data files. For about anything else, there's usually more than one way to read the data, and for some inputs one way works better than the other. Plus, there are options (such as extracting a specific region from an Excel sheet).

I'm not sure what to do in the context of ProjectTemplate. For each new data file added by the user, we could create a small `.R` file that contains code that loads the file, just like it's loaded now. The convention would be that for each data file there must be a corresponding `.R` file. @KentonWhite: What do you think?

Back to this pull request: If all data are read by R scripts, then the output of these R scripts defines the new variable names. These scripts can be sourced in a new environment, and whatever ends in this environment are the new variables.

KentonWhite: `data.files.loaded` is returned and loaded into `my.project.info$data`. `my.project.info$data` is then used in `.convert.to.data.table` to find loaded data frames and convert to data tables. Would it be possible to put the logic in `.convert.to.data.table`? Could we check for *any* data frame  that begins with the loaded variable name. This would catch `myvar.sheet1` as well as `myvar.myworksheet` and any other data frames created with a reader returning multiple datasets. It also fences in this change to only projects converting to data tables -- those projects that would most benefit from this change.

Comments and thoughts?

Hugovdberg:
I think this is fixed by a combination of #160 as that includes the function `.var.diff.from` and the proposed #187 as that moves the check of which variables were loaded before the conversion to `data.table`. This way only exactly those variables that are loaded are converted to `data.table`.

KentonWhite:
Thanks @Hugovdberg.  Do you think we can close this PR in light of #160 and #187? We can always reopen if needed...

Hugovdberg:
I think this could be closed for now, to me it seems the way #187 handles it is more stable as it handles all cases where the variable name assigned is different from the "suggested" value, not just for xls(x).

KentonWhite:
Closed for now. Will reopen if #187 doesn't address this issue adequately.

PR #143, opened 2016-01-18
add dplyr to default libraries in global.dcf

pavopax:
suggest adding dplyr as a default library to load

### DIFF
--- a/inst/defaults/full/config/global.dcf
+++ b/inst/defaults/full/config/global.dcf
@@ -5,7 +5,7 @@ recursive_loading: FALSE
 munging: TRUE
 logging: FALSE
 load_libraries: FALSE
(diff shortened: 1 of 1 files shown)


KentonWhite:
Excellent idea and thanks for the PR. @krlmlr before I merge, can you think of any reason adding dplyr might break backward compatibility?

pavopax:
Hopefully not but unfortunately i don't have extensive R experience. 

One thing to note though is that some 'dplyr' functions have similar names to 'plyr' functions so it's standard to first load 'plyr' then 'dplyr'

krlmlr:
This is precisely where backward compatibility breaks. That said, it affects only new projects, which supposedly adapt the list of their loaded libraries anyway. I don't have any particular preference here.

PR #146, opened 2016-06-10
trying installing missing packages

wilmeragsgh:
Installing and loading automatically the missing packages specified in config/global.dcf lib section.

### DIFF
--- a/R/require.package.R
+++ b/R/require.package.R
@@ -42,9 +42,9 @@ require.package <- function(package.name, attach = TRUE)
     #function.name <- deparse(sys.calls()[[sys.nframe()-1]], nlines = 1)
     message('We load or install the required libraries')
     #stop(paste(function.name, ' requires package ', package.name, '.\nPlease install ', package.name, ' by running install.packages("', package.name, '") and then try re-running load.project()', sep = ''), call. = FALSE)
+    require(package.name,character.only = T)
   }
 
-  library(package.name,character.only = T)
   invisible(NULL)
 }
 
--- a/R/require.package.R
+++ b/R/require.package.R
@@ -38,10 +38,13 @@ require.package <- function(package.name, attach = TRUE)
   }
 
   if (!success) {
-    function.name <- deparse(sys.calls()[[sys.nframe()-1]], nlines = 1)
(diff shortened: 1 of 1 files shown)


KentonWhite:
Thanks for the PR @wilmeragsgh! I'll have a look at it over the weekend. I see that the Travis CI build is failing for the PR. Can you see what is causing the failure please?

wilmeragsgh:
Hey @johnmyleswithe didn't find the bug but please the feature that i
mention to this already-great project, greatings!

2016-06-10 9:19 GMT-04:00 Kenton White <notifications@github.com>:

> Thanks for the PR @wilmeragsgh <https://github.com/wilmeragsgh>! I'll
> have a look at it over the weekend. I see that the Travis CI build is
> failing for the PR. Can you see what is causing the failure please?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/johnmyleswhite/ProjectTemplate/pull/146#issuecomment-225179744>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AKRANmQCl-HhSBAAL5GA1dgBKpKLyWN2ks5qKWRvgaJpZM4Iys29>
> .
>



-- 
*Wilmer González.                                       *
ve.linkedin.com/in/wilmeragslin <http://ve.linkedin.com/in/>

KentonWhite:
I like the PR! Thanks so much. I think that TravisCI is complaining about the use of `library` instead of `require`. Using `library` to load the package will cause problems with CRAN. Can you use `require` to load the package instead of `library` please.

The library call should only be executed if there is no success requiring the namespace of attaching the package. As the code is written, the package may be attached twice. While this shouldn't cause any difficulties, it is perhaps better if the package is only attached once.

Also can you use the same way of attaching the packages as above the code changes:

```
   if (attach) {
     success <- require(package.name, character.only = TRUE)
   } else {
     success <- requireNamespace(package.name)
    }
```
This way the package is either attached or just the namespace required if needed. If you can, it would be great if this code is only written once. Perhaps if feeling ambitious extract the code that attaches or requires the namespace into its own function!

Please let me know if you can make these changes. Thanks again for the PR!

wilmeragsgh:
Hey @johmyleswhite i did the changes check it out and let me know!

Issue #147, opened 2016-06-21
more flexible caching?


eribul:
I would suggest to allow some flexibility for cachnig (just to pass a `...`to save would be a atsrt). I sometimes like to pass argument `compress = FALSE` when I work with large data sets and it takes a lot of time to compress/decompress.
I might also prefer `saveRDS` to `save`.

KentonWhite:
I hadn't really thought about the choice of defaults for caching. It does make sense to allow some flexibility. I think passing in arguments to `cache` would be the easiest (rather than controlling the default behavior through the global.config file).

Is this a feature you might be willing to try implementing?

connectedblue:
I have proposed a change #152 which enables objects in munge scripts to be cached only if they don't already exist.  This saves time during subsequent load.project() calls, especially for large transformations that can take a while to cache and usually you haven't made any changes.

If you have created a dataframe called `foo` from the raw datasets, you can have the command:
`cache("foo", always=FALSE)`
in a script which will save the `foo` object the first time, but will skip the caching step in subsequent runs.  Of course, if you change the logic that creates `foo`, you will want to refresh the cache.  In this case, the call
`clear.cache("foo")`
will delete that item from the cache, forcing the `foo` variable to be re-cached next time load.project() is run.

KentonWhite:
I like this a lot and think it is a great step toward more flexible caching. Really like the addition of `clear.cache`. It is so much safer than manually deleting cached files. 

@connectedblue what do you think about adding a helper function `is.cached` that can be used to skip munging steps completely if the object exists in the cache? 

```
if(!is.cached("foo")) {
  # Expensive munging operation
}

connectedblue:
Yes, the `is.cached()` will be very helpful.  I'll add that into the branch.

connectedblue:
Actually, I think there is more to this than just this function.  The user still has to write the boiler code to test if it exists, do the processing and then cache with `always=FALSE`.

I think there is a further helper function that can do all that, plus emit warnings if the lines are skipped because the item is already cached, which would be helpful to remind if the user forgot to clear the cache when they changed some munge logic for that variable.

Something like:

`
dynamic.cache(var, {
      # Expensive munging operations
})
`

Again, I'll work up some code to demonstrate the context.  With an interface like this, you can start to do some smart caching functions like re-run the munge logic if the hash of the code block has changed compared to the value in the cache (as discussed in #80)

KentonWhite:
[...]

PR #148, opened 2016-06-22
allow extra arguments to be passed from cache to save

eribul:
Suggestion as discussed in #147.

NOTE: I have only updated the R-file! I have a newer version of Roxygen2, that uses a slightly different approach to logg its version number in the generated files etc. I therfore chosed not to to update since it might introduce bigger changes than intended!

NOTE ALSO: I do not pass all the testthat tests after the change but I do not think this is related to the suggested change but to something else (that I do not have the possibility to investigate right now). Hence, I do not think that it was this change that introduced the test failure on my computer.

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -8,6 +8,8 @@
 #' @param variable A character vector containing the name of the variable to
 #'  be saved.
 #'
+#' @param ... additional arguments passed to \code{\link{save}}
+#'
 #' @return No value is returned; this function is called for its side effects.
 #'
 #' @export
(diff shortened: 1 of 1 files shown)


KentonWhite:
Thanks for the PR! I see what the problem is that is causing the build to fail. Since you didn't update the docs there is mismatch between the docs in the R files and the man pages. I'll fix that up when completing the code review.

KentonWhite:
Looks good. Will fix up the documentation with Oxygen.

Issue #99, opened 2014-09-09
Feature Request: Create file templates


biocyberman:
I would like to put a request on a feature that allows creating various file templates:

Currently the project configuration file `config/global.dcf` file is good to load the project as a whole and any scripts therein will not have to handle library import and data preparation. But if we just one to run one script in with `R --vanilla` for example, this would not work because libraries and data are missing. My suggestion is to implement a function like `create.file.template(name = "file_name", dir = "munge", type = c("R", "Rmd")[1],   library = c("dplyr", "lubridate", "ggplot2"), data =  "my_dataset1")`. The function then create a file with appropriate content template based on input parameters.   For example, with the aforementioned command, it would create an R file under `munge` directory with the beginning of the file like so:

``` r
# Project: My Test Project
# File: file_name.R
# Date Created: today()
# Author: TheAuthor
# Email: dummy@email.com
# Anything else from global.dcf

library(dplyr)
library(lubridate)
library(ggplot2)

my_dataset1 <- read.csv("../data/my_dataset1.csv")

```

This way the file is both consistent with project structure and self-contained. 

Would this be something worth doing?

Issue #80, opened 2014-04-10
Provide an easy way to invalidate cache


krlmlr:
Currently, file has to be removed manually from `cache` subdirectory. Can we do better?

```
invalidate.cache <- function(names = NULL) { ... }
```

Or even a config option?

connectedblue:
My proposal #152 contains a function to clear the cache of one or more objects:
`clear.cache(c("foo", "bar")`
will delete those variables from the cache if they exist.

It provides an interface and a simple implementation behind (just deleting the correct files from the cache directory).  If more advanced stuff comes along such as cache aging, maybe this interface could be extended to incorporate those?

PR #152, opened 2016-09-10
Some updates to caching and config

connectedblue:
Hi John

I love using Project Template.   I love how it brings clarity to the analysis process.

For a while I've been creating some custom functions in lib to help certain projects, and I've been cut and pasting them into each project which is just horrible.  So I created a fork of the package to store the functions in my own version of ProjectTemplate.

I thought I'd share them with you to see if they may be useful to put in the core package.  The functions I added are:

- update cache only if there's not already a cache object (but do default behaviour of caching always).  This allows munge scripts to skip caching if it takes a long time which allows load.project to run quicker.

- have a file called custom.dcf in the config directory which adds config variables which are relevant for that project only.  These can be anything (char, numeric, logical) etc and allows for analysis configuration (e.g. I use it to determine which variables should be cached in the analysis)

 - new function clear.cache which allows specific cache objects to be deleted (forcing data to be re-read from raw data).  I also use this to only run long analysis in munge if the data isn't already cached which saves time again when load.project is run.

The overall theme of the changes are about trying to write munge processing in a way that executes quickly when load.project is run.  If you are building up specific dataset transformations on the raw data, you can cache them if they take a long time to run.  The munge logic can then be written to only execute if the object doesn't already exist in the global environment

If however your data changes, or you update the transformation logic, then you simply clear the cache and run load.project again.

Appreciate this is a long note, but would appreciate feedback on whether you feel this functionality would be useful to develop into the core package.

best regards
Chris


KentonWhite:
Hi @connectedblue and thanks for the PR.  Sorry I haven't responded sooner and I'm looking forward to digging into the PR. Just a heads up that it will take me a few days to review and have some well thought out comments.

Initial impressions are this is a great direction. The munge step with its all or nothing approach can be frustrating. Adding some additional logic into the munging could be very helpful. 

There have been some discussions in the issues already on the features you are adding. I'm thinking specifically of issue #99, #147 and #80. Would appreciate your thoughts on how your suggestions here solve the issues discussed or are not relevant.

connectedblue:
Hi @KentonWhite, thanks for the reply.  Look forward to seeing your comments.

I added some comments to #147 and #80 which are partially addressed by this PR.

Issue #99 is interesting, but I don't think immediately related to a custom project config.  However,  I have developed a flexible templating architecture for projects, reports and src files which I will submit in a future PR.

R/load.project.R
```diff
@@ -275,7 +275,25 @@ load.project <- function(override.config = NULL)
...
                               .boolean.cfg)
-
+  
+  # add some project config if it exists in a file called custom.dcf
+  custom_cfg <- file.path('config', 'custom.dcf') 
+  if (file.exists(custom_cfg)) {
+        custom_cfg <- translate.dcf(custom_cfg)
+ 
+        # all items are char, so convert them if they look like logical or numeric
+        convert_type <- function(v){
```
KentonWhite: I'm wondering if there is a way to replace `convert_type` with `.boolean.cfg`. Even if there isn't, can you try to use the `.normalize.config` idiom above? It does exactly what you want to do here but in a reusable way!

KentonWhite:
Here are my thoughts on the `custom.dcf` file.

First I like the idea of having config variables in a single place that is loaded before the rest of the project. I also like having a separate config file from `globe.dcf` to put these in. I find a real need to add project specific global variables from time to time, so this potentially solves a real problem.

I'm curious about your thoughts on the following design point. How is having a `custom.dcf` file different from creating a file `data/globals.R` or `lib/globals.R` which define global variables? With `config.dcf` we are restricted to text only values (or values that can be coerced to logicals). This restriction might require some extra work downstream to convert the text field to R objects. It could also lead to code duplication where several munging scripts must do the same conversion. A `globals.R` file  would allow the values to be written in idiomatic R and not require further conversion downstream.

I'm really curious your thoughts! I love having config values in one place and like the `custom.dcf` approach. I also need to weight this improvement against other options available.

R/load.project.R
connectedblue: Yes, I looked at the .boolean.cfg and it did look very similar.  I agree it would be more re-usable to follow the paradigm, rather than the messy inline function I used here.  I'll tidy up in my branch and put out again for review.

connectedblue:
Yes, this is a valid point.  The case for having a custom dcf is that it is "ProjectTemplate-like" rather than just  "R-like".  You get used to changing your config in the directory and it follows the same format.  You can also add project specific config safely to the config object (admittedly you can do this in a globals.R file as well).  

I think it promotes good practice to have config stored under one place, otherwise people end up creating random project variables like "keep_raw=TRUE" in their munge scripts and it's not immediately clear that;'s a config variable that can be tuned by collaborators.

It's also easier to add comments in if it's an R file.  One thing I was going to add to the DCF reader was to strip out comment lines before reading, but that's quite complicated due to having to create an intermediate temporary file for processing.

On balance, what about the following:

 - create a skeleton template in /lib/globals.R with a comment in to signpost and encourage people to put project specific content in there.
 - create a create.config(var=foo)  function that safely adds var to the config object (so if the way that config is handled in the future, scripts won't break, the function can be updated to be compatible).

The `create.config` interface could be formatted in the globals.R file to look like a config file:
`create.config(
         keep_raw=TRUE,       # keep the raw data.frames in memory
         title="My Analysis",   # string to output in all report headings
         num_loops=34          # number of times to try the data loop process
)`
This pretty much gives everything that is in `custom.dcf` out the box (with = instead of colon).

The comments format works too in the middle of function calls which is a bonus.

What about this alternative approach?  I can test out in one of my branches and re-submit?

connectedblue:
Added some changes to remove the concept of `custom.dcf` and created a new function instead called `create.config` to load config from `lib/globals.R`

connectedblue:
And just uploaded a new is.cached function

connectedblue:
Thinking about this overnight, I think the function should be called `add.config()` because that is what it is doing.  It's updating, not creating the `config` object.

It would also be natural if some-one wanted to over-ride a project specific config in an analysis, they can just put another `add.config()` command to the top of that `src` script and it would replace whatever config had been set in `globals.R`

KentonWhite:
This is shaping up rather nicely. I'm assuming that we would have to document how to access the config variables in the documentation, e.g. `config$my_config_variable`. I like that it isn't cluttering the global namespace, which is a good idea.

One minor detail now, is thinking if it is necessary to load `global.R` before the other files in `lib`. I can imagine that a user might reference one of the project config variables in a helper file that might be loaded before `global.R`. 

On one hand `lib` files should only contain functions, in which case the load order wouldn't matter since functions are lazy loaded in R. On the other hand,  a user _might_ have some scripts that run when a `lib` file is loaded (not recommended of course!) that could be expecting the config vars to be defined. 

What are your thoughts?

connectedblue:
Yes, it occurred to me also that lib loading order could matter.  You could argue that users should understand the dependencies in their own code, but there again, they might also assume that `ProjectTemplate` was doing some background magic and the order doesn't matter.

A simple fix is to prepend a 00- in front of the file name to ensure it is always called first (unless the user intentionally places something before it).

Of course, we are breaking our own rule if lib should only contain function definitions, since we are inserting an executable statement.  However, I think `lib` is a more natural place for this file than say `data` (or even `config`) so I think it's an exception worth making.

I've updated the branch to make the documentation clearer on how to access the config as @KentonWhite  suggested.

connectedblue:
@KentonWhite One thing I've learned is not to submit too many features in one pull request!

I think this PR is complete as it stands.  I'm going to open another one to deal with the `dynamic.cache` proposal referenced in #147.  That is a complicated update, so there's no benefit in mixing it with this PR.

Are there any other changes you think need to be made to this PR before it can be merged?

KentonWhite:
I was going to ask you to separate the two features but didn't want to scare you too much :)

I'll take a detailed look at the PR overnight / tomorrow morning. I think if there are any changes they should be minor.  One thought on ensuring the `globals.R` loads first. Since ProjectTemplate embraces convention over configuration, it would be appropriate to hard code in the library loader that `globals.R` is loaded first. What do you think?

connectedblue:
Yes, if it's acceptable to encode convention, then we should allow `load.project` to do magic in that case.

Small patch added to the branch to achieve this, which I've tested on one of my more complex lib directories.

KentonWhite:
Thanks! Really like how the `add.config` feature turned out -- thanks so much @connectedblue :)

A few housekeeping tasks before merging:

- Can you pull out the caching code and only have the config code in this PR please. The cache code should be in a separate PR

- Can you squash all of the commits into a single commit please? This makes it easier to keep everything up to date

- Can you update the additional documentation for the website and readmes please? I'm thinking of `website/architecture.markdown` and `inst/defaults/full/lib/README.md`. There might be others, but these are the key ones to update.

connectedblue:
Hi @KentonWhite .  I've made those final changes as you suggest.  This PR only contains the `add.config` functionality and documentation.  A couple of final queries from me:

* I changed the architecture and configuring markdown documents, but wasn't able to view the HTML.  Is that something you generate to update the site?

* From reading the documentation, github squashing looks like something you do when merging the commit from the PR, rather than me.  Is that right?  If I have to do it, how would I go about that?

connectedblue:
Thanks @KentonWhite.  Just working on another feature for smarter caching which I'll post for review in a seperate PR

PR #155, opened 2016-09-20
Faster cache function

connectedblue:
This PR proposes an update to the `cache()` function and introduces a new function `clear.cache()`.  Together these changes give the user a faster experience when processing large datasets when `load.project()` is called.

It is a large update, and I have tested the scenarios on a variety of my own projects.  I have also made it backward compatible, so existing scripts will still work (but they will work faster).

The main change is the rules which are applied when the cache function is called, so that a cached object is only written out if it has changed.  There are two main ways to use it:

```
cache("foo")
```

This will create a cached object of the `foo` object in the global environment the first time it is encountered in a `munge` script.  If the object has not changed in subsequent invocations of `load.project()`, it will not be cached again.  Messages are output to tell the user what is going on.

The second way is as follows:

```
cache("foo", CODE={
     x<-loaded_data$name
     x<-as.character(x)
     x[[1]]
})
```

Here the `cache` function is "wrapped around" some code in the munge script directly.   In this case, the   `CODE` is executed in the global environment (so a variable x is created) and also the variable `foo` is assigned the result of the execution, also in the global environment (in this case the first entry of the vector `x`).  `foo` is then written to the cache.

The cool thing about subsequent invocations of `load.project()` is that if the code has changed since `foo` was cached, it will be evaluated and re-cached again.  If the code hasn't changed, it will not be re-evaluated which saves a lot of munge time.

There is a final variant of this as follows:

```
cache("foo", depends=c("x", "y"), CODE={
     x<-loaded_data$name
     x<-as.character(x)
     x[[1]]
})
```

In this case, if either the code, or the variables `x` or `y` have changed, then the code is re-executed.  This is important if the code chunk references variables where you want to re-evaluate if those other variables have changed.

I tried this on a project where my munge processing was taking several minutes the first time.  In subsequent runs, the `load.project()` ran almost instanteously when there were no changes to the cached code or datasets.

Note that the loading of data from the cache has not been touched in this PR.

Finally, a `clear.cache` function is provided to allow the user fine grain control over the cache - they can remove objects from the cache if they want to force a re-run of all data from scratch.  this function takes the form:

```
clear.cache("y", "z", "a")
```
or
```
clear.cache()
```
to clear everything.

Sorry for the long description.  This is a big update, but I hope is a useful addition to speed up the user experience during large analysis.  This also addresses the issue raised in #147.

*** Note I edited this example from the first version of this post to show the `CODE` parameter accepting `{ .... }` instead of wrapping quotation marks.


connectedblue:
Hi @KentonWhite .

Sorry to push an update while you're reviewing.  I just improved the user messaging when cache's are being performed.  I had good messaging when it was skipped, but not otherwise!

Now when you run load.project, you can see exactly what caching steps are being performed and why.

I haven't changed the functionality or core logic since the PR, just these messages.

Look forward to seeing your review and suggestions.

wlandau:
@connectedblue The work you did on caching sounds great. I like that the `cache()` function can track dependencies.

This may be a bit late for me to say, but what are your thoughts on [storr](https://CRAN.R-project.org/package=storr)? It's totally focused on caching, and it's separate from the issue of resolving dependencies.

connectedblue:
Hi @wlandau 

Yes, `storr` looks a better solution for the underlying cache and hash operations.   If I had known about this beforehand, I probably wouldn't have tried to roll my own caching.
 
It looks like it would be one big big file in the cache directory with all the cached variables in it.  I would still need the logic for code evaluation and dependency tracking because those are solving specific `load.project()` issues to speed up when it is called over and over.  

A rough impact on my code would be that the first part of logic switching still stays, because we want to generate specific messages to inform the user what is going on (just tweaks to use the correct `storr` methods).  However most of the helper functions from `.create.cache.hash` downwards would be redundant and would simplify the function.

Obviously the loading from the cache parts of `load.project()` would also need to be changed to use these methods.

So, my thoughts at this stage is to leave the code as in the PR because it does the specific job it sets out to.  However, if it emerges that we need to have more sophisticated caching, then a re-write would be worth it in the future.  One such driver which has crossed my mind often is to unify the way `knitr` and `ProjectTemplate` do caching because at the moment they duplicate which is a potential cause for error in reproducibility.  That's a big task though requiring co-operation between the two packages.  I don't think I have the energy to embark on such a mission - I prefer to spend my time on data science :)

If it turns out that my function breaks horribly during testing on other people's projects, then I'll look at a total re-write.

Would you mind installing my `ProjectTemplate` from my github branch `fastcache` and test caching  on some of your own projects that have large datasets?  I'm interested in feedback.

wlandau:
@connectedblue I'm brand new to ProjectTemplate, so I used the example from the ["Getting Started"](http://projecttemplate.net/getting_started.html) page. My version of `muge/01-A.R` looks like this.

```r
cache("first.letter.counts", depends=c("letters"), 
  CODE="ddply(letters, c('FirstLetter'), nrow)")
cache("second.letter.counts", depends=c("letters"),
  CODE="ddply(letters, c('FirstLetter'), nrow)")
```

I ran `load.project()` and saw `first.letter.counts` and `second.letter.counts` automatically appear in `cache/`. Subsequent calls to `load.project()` responded as expected to changes (and lack thereof) in `CODE`. I really like that caching and dependency tracking are integrated into the user's code without the need for a formal overarching [Makefile](http://robjhyndman.com/hyndsight/makefiles/) or [remake.yml file](https://github.com/richfitz/remake). This is convenient and helpful for projects where interactivity is front and center.

However, I am concerned about what happened when I replaced the contents of `munge/01-A.R` with the following.

```r
cache("first.letter.counts", depends="MISSING", 
  CODE="ddply(letters, c('FirstLetter'), nrow)")
```

When I ran `clear.cache(); load.project()` in a new R session, I was surprised that `cache` did not complain. I thought it would throw an error since `MISSING` did not exist.

In addition, I have a couple ideas for future work. First, it would be really convenient if `CODE` could be a code block rather than a character string. (When I removed the quotes from `CODE` just now, I got the wrong answer for `first.letter.counts`.) This may not require much extra work. For example, you could first convert `CODE` into a string and then go about your business. In [another package](https://github.com/wlandau/remakeGenerator), which I used as the workflow manager for my most recent projects, I have a function called `strings()` that you're welcome to copy. (I don't recommend importing it in the NAMESPACE because my package is not on CRAN yet.)

```r
library(devtools)
install_github("wlandau/remakeGenerator")
library(remakeGenerator)
txt <- strings({
  x <- 1
  y <- 2
})
eval(parse(text = txt))
print(x)
print(y)
```

Looking far into the long term, you might write a version of `cache()` that resolves targets (objects to cache) and dependencies just by parsing a block of R code. I think this would be more natural and less error-prone than having to use the `variable` or `depends` arguments. So far, I haven't seen this totally implemented. The [remake](https://github.com/richfitz/remake) package seems to come closest, but the framework is somewhat rigid.

connectedblue:
HI @wlandau 

Thanks for downloading and running the package.  It's great to get feedback on use cases - you have got me thinking.  

Your first point:  the cache didn't complain when a dependent variable MISSING didn't exist.  I coded this as a feature so that cache wouldn't stop unnecessarily.  Of course, if you had referred to MISSING in the `CODE` block, it would have course of complained because it wouldn't be evaluated.

However, your comment that you like the `depends` and `variable` right there in the code got me thinking:  the user knows what they want to cache so silently trying to continue when there is a mistake in the analysis is wrong.  I'm going to take that feature out, so it fails and the user can investigate.  

Your second observation about using a code block is what I actually started out implementing - I wanted that interface too (see my comment on #147 which was the beginnings of this whole function).  However there is a key problem with this approach.  When a code block is passed into a function, the first time it is referenced, it is also evaluated.  In this case, I needed to pass it into `digest` to calculate the hash to work out whether it should be evaluated - but of course it would have been evaluated anyway.

So, in order to get the fast functionality - only evaluate if the code changes - I needed to pass in as a character string and calculate the hash on the string.  In my own work, I have some munging functions that take several minutes to process.  So I don't want to wait each time I run `load.project()`.

It's an interesting idea to determine the dependencies automatically from the code block.  This may not always be possible (quite often I run custom functions from `lib` in my munge scripts, so you would have to inspect those as well for dependencies).  It is something I will bear in mind for future development, but for this iteration, I wanted to get a simple to use, fast and smart cache.

I'll take a look at your package - there's always lots to learn from how others solve these problems.  Thank you so much for taking the time to review my change - I'll go and make the update now and release another version with the silent dependency drop removed.

connectedblue:
Pushed a fix for the issue @wlandau found - `cache()` will stop with an error if a variable does not exist in the global environment.

wlandau:
@connectedblue I see your point about the evaluation of code blocks. In fact, I recently struggled with it in [downsize](https://CRAN.R-project.org/package=downsize/vignettes/downsize.html). There, it was not a show-stopper because I did not need to know anything about the code block arguments before deciding which one to evaluate.

On the other hand, for reasons that I do not understand, `strings()` does not seem to have this problem.

```r
> library(remakeGenerator)
> ls()
character(0)
> args = strings(variable = my_data, depends = some_dependency, 
+  CODE =  {tmp <- MISSING
+    my_data <- tmp})
> ls()
[1] "args"
> print(args)
                                      variable 
                                     "my_data" 
                                       depends 
                             "some_dependency" 
                                          CODE 
"{\n    tmp <- MISSING\n    my_data <- tmp\n}" 
> print(MISSING)
Error in print(MISSING) : object 'MISSING' not found
> print(tmp)
Error in print(tmp) : object 'tmp' not found
> print(my_data)
Error in print(my_data) : object 'my_data' not found
```

So maybe if you rip the guts out of `strings()`, put them at the beginning of `cache()`, and add some argument checking, there may be hope after all.

Thanks for the interest in [remakeGenerator](https://github.com/wlandau/remakeGenerator). That toolkit (together with [downsize](https://CRAN.R-project.org/package=downsize/vignettes/downsize.html) and [parallelRemake](https://github.com/wlandau/parallelRemake)) focuses on long jobs at the expense of interactivity (one of `ProjectTemplate`'s big strengths, as I see it), so I think we are complementing each other rather than competing.

wlandau:
@connectedblue Also, it just occurred to me that you might want to send `CODE` through `formatR::tidy_source(text = CODE, comment = F)` initially so that changes to whitespace and comments do not trigger unnecessary updates to cache entries. We have [remake](https://github.com/richfitz/remake) to thank for that idea.

connectedblue:
Thanks @wlandau, that's a great idea.  I'll add that in as an extra convenience for the user.

I'll experiment with `strings` a bit more to see if it helps use code blocks here.  What do you see as the advantage of passing a block rather than a string (apart from that it's more R-like)?

I'm working on another feature for `ProjectTemplate` that allows the user to set different preconfigured boiler plates for different type of projects (e.g. `shiny` projects,  `knitr` reports etc).  It could be interesting to see if there's a way to have a `remake` template  to build a bridge between the two packages.  I'll drop you a note when I post the PR and we can discuss further.

wlandau:
Sure @connectedblue, glad to help.

As I see it, the main advantage of passing an unquoted code block is user convenience. With `CODE` passed as a character string, any inner quotes that match the outer quotes will need to be escaped. It can be annoying and error-prone, syntax highlighting in editors goes away, and the user can't always easily copy and paste `CODE` into an interactive R session for informal debugging and testing.

I agree, it would be awesome to bridge `ProjectTemplate` and [`remake`](https://github.com/richfitz/remake). Hopefully there's a way.

connectedblue:
@wlandau - very good point.  Let me see if I can get that working with blocks.  As I said, I had trouble when I started down that route, but maybe I can learn from your work.

I've pushed another update adding new messaging describing the current cache status.  You simply type `cache()` with no arguments, and the status and dependencies are parsed for you.

wlandau:
@connectedblue Neat! The overall summary from `cache()` with no arguments is a helpful reference. Plus, it's kind of like generating a Makefile. Long term, if you wanted to run with that in the future, it might be possible to
- Generate a real Makefile, which would open up parallel computing with `make -j`.
- Generate a `remake.yml` file for integration with [`remake`](https://github.com/richfitz/remake).

connectedblue:
OK, thanks to some inspiration from @wlandau, I have adapted the `CODE` parameter to accept code blocks instead of a string.  The functionality remains the same - CODE is only evaluated if the actual code (minus comments and whitespace) has changed since last time.

This means that any kind of valid R code can be passed into `CODE` for evaluation and caching.

I feel that this function is pretty complete now, and ready for review.  I won't add any more to it now except in response to comments from others.

wlandau:
@connectedblue Ran just fine for me in a test. Thanks for your work on this.

KentonWhite:
@wlandau Thanks so much for testing and feedback of this PR! The community is just Awesome :)

@connectedblue Do you think this PR is pretty stable now? I didn't want to jump into reviewing while this great conversation was happening! If it is pretty stable I'll review and merge pretty quickly as I'm really excited for this feature.

One small question, if it possible to optionally drop the variable name `CODE` in the block `CODE={...}`? E.g.

```
cache("foo", {
    # Code block here
  }, depends = c("x", "y")
)
```

connectedblue:
@KentonWhite , yes I agree - the community is very awesome.  

I swapped around `CODE` and `depends` to allow for dropping the name as you suggest.  This supports the form you show above, but because of variable ordering it also supports:

```
cache("foo", , depends = c("x", "y"), {
    # Code block here
  })
```

which I think is neater since the `depends` doesn't get lost at the bottom or confused with the code block itself.  So, you have the best of both worlds depending on your style preference.

I think the feature is stable now.  I have added some documentation for the website - a new page on caching and also a new sidebar menu item.  I can't test the new documentation, but hopefully it compiles OK to show the new page.

Look forward to your review comments.

connectedblue:
Hi @KentonWhite 

I'm starting to learn about the importance of regression testing (on #157 )!  I haven't written scripts before using testthat, but they seem quite simple.

I shall create some tests for this feature given it's complexity - it'll also help ensure it doesn't break with future changes.

I'll let you know when the test scripts are pushed.

connectedblue:
@KentonWhite  - I've now written a package of tests to go through all the various caching cases:

```
test_that('caching a variable that doesnt exist fails with correct message'
test_that('caching a variable not already in cache caches correctly'
test_that('caching a variable created from CODE caches correctly'
test_that('re-caching is skipped when a cached variable hasnt changed'
test_that('re-caching is done again when a cached variable has changed'
test_that('re-caching fails with correct message if cached variable is not in global env'
test_that('re-caching a variable created from CODE only happens if code changes, not comments or white space'
```

The tests are quite comprehensive - they re-load in from cache and check against memory, and also check file modification times in the cache to check if caching has completed or skipped as appropriate.

I think this feature is complete now - I feel better that there are automated tests.  It was getting tricky manually testing everything.

Let me know if you think there's anything else to look at for this feature.

KentonWhite:
Awesome! I'll try to merge this weekend, Monday at the latest.

KentonWhite:
@connectedblue Thanks so much for this PR! Sorry it has taken me so long to review :(

connectedblue:
@KentonWhite Thanks very much - it was a fairly complex update so I'd expect it to take some time.

I'm interested now in getting feedback from any users who may be early adopters - I'm sure there are people who like to keep on the bleeding edge with `master`.  I'll keep an eye on the issues list and jump on anything that comes up.

PR #157, opened 2016-09-28
Cleaner exit when running load.project() in the wrong directory

connectedblue:
This PR is a small bug fix to produce cleaner behaviour if `load.project()` happens to be run in the wrong directory.

Currently, there is a confusing error message thrown, some variables are loaded into the global environment and two blank directories `data` and `cache` are created.  

This patch checks first to see if the current directory is indeed a `ProjectTemplate` project as determined by the presence of the `config` file, the `data`directory and  the `cache` directory.  These are the ones required by `load.project().

As a bonus, the code also checks when `create.project()` is run and if that happens to be a `ProjectTemplate` also, the function will exit with an explanatory error.  This prevents users from inadvertently creating nested projects which could get very confusing.

No documentation changes are required - the error messages are self contained.

There are a couple of internal (non-exported) helper functions `.project.info()` and `is.ProjectTemplate()` which could be of interest to authors of new features in the future


connectedblue:
The regression tests now pass.

They were previously failing on the migration test scripts because the test data didn't contain a valid structure that would pass the checks under `load.project()`.  To fix this, I added a `cache` and `data` subdirectory to the test data.  `git` does weird things with empty directories, so I put a dummy file in each to ensure they are checked out properly.

So I didn't change the test itself, just the test data.  The tests now pass which is good news.

While I'm at it, I will write an additional test to check that the new error messages introduced by this feature are displayed in the correct conditions.

So,  at this point in time, this PR is not ready for review while I write the new test script.

connectedblue:
Hi @KentonWhite .

This PR is ready for review now.  The feature is as described in the beginning, but I've added some tests into the regression suite to check that the messages display in the right place.

Look forward to your comments.

KentonWhite:
@connectedblue Thanks for this.

PR #161, opened 2016-10-17
Allow to call RSQLite's initExtension()

crubb:
RSQLite has the function initExtension() to automatically load extension-functions.c (50.96 KB) contributed by Liam Healy on 2010-02-06 15:45:07 (https://www.sqlite.org/contrib?orderby=date)

It provides mathematical and string extension functions for SQL queries using the loadable extensions mechanism. Math: acos, asin, atan, atn2, atan2, acosh, asinh, atanh, difference, degrees, radians, cos, sin, tan, cot, cosh, sinh, tanh, coth, exp, log, log10, power, sign, sqrt, square, ceil, floor, pi. String: replicate, charindex, leftstr, rightstr, ltrim, rtrim, trim, replace, reverse, proper, padl, padr, padc, strfilter. Aggregate: stdev, variance, mode, median, lower_quartile, upper_quartile.

The pull request adds a config option in the .sql files to call initExtension()

Thanks!

### DIFF
--- a/R/sql.reader.R
+++ b/R/sql.reader.R
@@ -174,6 +174,11 @@ sql.reader <- function(data.file, filename, variable.name)
 
     connection <- DBI::dbConnect(sqlite.driver,
                             dbname = database.info[['dbname']])
+
+    if (!is.null(database.info[['plugin']]) && database.info[['plugin']] == 'extension')
+    {
+      RSQLite::initExtension(connection)
+    }
   }
 
   if (database.info[['type']] == 'postgres')
--- a/R/sql.reader.R
+++ b/R/sql.reader.R
@@ -90,6 +90,17 @@
 #' dbname: herokudb
 #' query: select * from emp
 #'
+#' Example 11
+#' In this example RSQLite::initExtension() is automatically called on the established connection.
+#'
+#' Liam Healy has written extension-functions.c, which is available on http://www.sqlite.org/contrib.
+#' It provides mathematical and string extension functions for SQL queries using the loadable extensions mechanism.
+#'
+#' type: sqlite
+#' dbname: /path/to/sample_database
+#' plugin: extension
+#' query: SELECT *,STDEV(value1) FROM example_table
+#'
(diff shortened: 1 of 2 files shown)


KentonWhite:
@crubb Thanks so much for this PR. I'll try to review today or tomorrow. Think this is a great idea!

Quick question -- do you know if these extensions work with other databases and not just RSQLite?

crubb:
Hey,

extension-functions.c is unique to SQLite and therefore initExtension() is (unfortunately) unique to RSQLite.

Best regards,
crubb

KentonWhite:
Nice simple change. I like that the default handles the case where plugin is not listed. 

Since this is an advanced feature of RSQlite, I don't think that the website documentation needs to be updated. However, can you update the comments at the start of the file with an example and some documentation about \ extensions. Then can you please run ROxygen to include the example in the RDoc files. This will let someone looking for the extension be able to find it.

Thanks so much for you contribution!

crubb:
Thank you for the feedback, happy to oblige!

- Should I adapt an existing sqlite example or should I add a new example and increment the number of the following examples?
- Should I explain extensions in the forementioned examples or at the top of the file around [sql.reader.R#L14](https://github.com/johnmyleswhite/ProjectTemplate/blob/a82e4170ddffb9b9dc18f26c25866e14e74c8dfb/R/sql.reader.R#L14)?

Thank you for the clarifications!

KentonWhite:
@crubb I would add a separate example at the end. My thinking is this:

SQLite is a very popular introductory database. It is the first example in the RDoc and really meant to get a new user started quickly. Introducing an advanced concept like extensions in the first example increases the cognitive load of an already daunting learning curve.

However, a person who is very advanced and looking to use SQLite extensions is likely to be specifically looking for that information. They will more likely read to the end, where they will find the example!

crubb:
Just extended the PR with the requested changes.

KentonWhite:
@crubb Awesome! Thanks!

PR #163, opened 2016-11-03
Clean exit for more functions when current directory is not valid Project template

connectedblue:
This PR proposes extending the feature in #157 to the following functions:

- `cache()`
- `cache.project()`
- `clear.cache()`
- `reload.project()`
- `test.project()`
- `stub.tests()`

These will now exit cleanly with an error message if they are executed in a non ProjectTemplate directory.

In order to simplify the implementation, two helper functions are created:  `.stopifnotproject()` and `.stopifproject()`.  The `load.project()` and `create.project()` have been refactored to use these new functions.

These functions now issue a `stop()` rather than `return(NULL)` so the tests have been modified to account for the new behaviour.

Finally, since this function is used in a wide range of functions, the helpers have been moved away from the `load.project.R` to a new file called `stopifnotproject.R` for ease of reference.


connectedblue:
Not sure why the travis tests all failed.  They all work locally on my machine.

I'll try and force a re-run in case there's a problem with the travis tool

connectedblue:
ok, there's something odd going on here.  I'll close this PR and investigate further.

connectedblue:
@KentonWhite This PR is ready for review now.

Sorry about all the closing and re-opening.  I finally tracked down the problem with a missing function which I've now added back.  What I can't explain though is how all my tests passed locally without this.  I hate it when things can't be explained ....

Anyway, all seems fine now - this PR just extends the earlier functionality across a lot more functions.

PR #166, opened 2016-11-18
Fix bug in .load.cache() and refactor

connectedblue:
This PR fixes a bug in cache loading:

If a variable containing an underscore is cached (e.g. en_docs), then on a subsequent `load.project()`, the `.load.cache()` function tests for the existence of a variable called `en.docs`, doesn't find it and then proceeds to load `en_docs` from the cache..  So the correct variable ends up in global env, but an unnecessary load has occurred if it was already there.

On inspecting the code, a few things became apparent from the way `.load.cache()` works:

 - the cache load code is essentially a cut and paste of the generic `data` load code.  It essentially supports loading any kind of cached variable extension, whereas the `cache()` function only ever writes `.RData` files (this was true even before the recent enhancements to `cache()`.  So, to be consistent, caching should only read `.RData` files and users should be discouraged from putting other types of data in that directory (that's what `data` is for after all).
 - The `clean.variable.name()` function is called before checking against memory.  This is where `en_docs` would get transposed into `en.docs`.  However, the `rdata.reader` doesn't use this cleaned variable name which is why the correct name got loaded into memory

I think that `clean.variable.name` needs to stay for raw data reads, but once a variable is in memory (maybe even created in a subsequent munge script) then it should be cached without modification.  This happens today by default (except for the edge case described above) but only because the `rdata.reader` doesn't enforce cleaned up variable names.

In fixing this, I have done the following:

- removed all the dispatch extension checking in favour of only looking for RData files
 - removed the clean.variable call
 - moved the whole function to the `cache.R` file to make it easier to maintain going forward (e.g. if the way caching is done changes going forward, the `.load.cache` and `.write.cache` functions are next to each other
 - removed the call to `rdata.reader` which only wraps a `load` call.  This is in case it is updated in the future to support `clean.variable.name` during data load.

No user documentation changes have been made as the overall behaviour hasn't changed.

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -158,6 +158,13 @@ cache <- function(variable=NULL, CODE=NULL, depends=NULL,  ...)
 }
 
 
+# Cache directory and extension used
+
+.cache.dir <- 'cache'
+.cache.file.ext <- '.RData'
+
+# Function to save to cache
+
 .write.cache <- function(cache.hash, ...){
         # cache.hash is a data frame with two columns:  variable and hash.
         # Row name VAR is the name of the variable to save.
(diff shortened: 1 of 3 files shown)


KentonWhite:
@connectedblue quite nice! Thanks for this.

PR #167, opened 2016-11-20
New clear() function to manage global environment

connectedblue:
This PR proposes a new user function `clear()` to provide some useful tools for the `ProjectTemplate` user to manage their analysis workflow.

The motivation behind this is because some large datasets (e.g. raw or derived datasets) may not change all that often during analysis, but other variables do (e.g. models, frequency summaries etc).  So a user may need to delete old variables and rerun `(re)load.project()` quite often but doesn't want to keep waiting for these large datasets to load from `cache` or `data`.

This function provides a new config item `sticky_variables` which can be configured with a comma separated list of variables that the user wants to keep in memory.  So if the `clear()` command is issued, everything in global env will be deleted except those configured things.

The user can also specify ad-hoc things to keep without having to edit `config/globals.dcf` by using a parameter option:  `clear(keep=my_var)`. 

If the user wants to really clear everything, they can issue a force:  `clear(force=TRUE)`.

As a bonus, there is a use case that works anywhere, not just  `ProjectTemplate`:  Typing `clear()` is so much easier than `rm(list=ls())`.

I've updated the website and manual documentation as well as including some tests for all the cases.


KentonWhite:
Thank @connectedblue! Should there be a `migrate` for this new feature since a new config var is being added?

connectedblue:
Hmmm ... good point @KentonWhite.  Let me think about this.  I'll play around with some scenarios.  I tested that migration worked ok on my project, but I'm not sure if migration specific config is required.

I'll get back to you.

connectedblue:
OK, I've checked what happens on when `migrate.project()` is run on on a previous project.

The following message is shown at the terminal:

```
Migrating project configuration
Your existing project configuration in globals.dcf does not contain up to
date configuration settings in this version 0.7 of ProjectTemplate.  They will
be added automatically during migration, but you should review afterward.
Your config.dcf file was missing entries and defaults
have been used.  The missing entries are:
 sticky_variables
```

This is what happens without any particular change for this file - the function picks up that there is a new config item automatically.

We don't need to set the default to anything different for migrated project, so I don't think there's a need for a custom migration behaviour.

KentonWhite:
@connectedblue thanks for checking on that.

PR #174, opened 2016-12-17
New feature:  custom templates

connectedblue:
This PR proposes a major new feature into `ProjectTemplate` which is the ability for the user to have their own custom templates applied on top of the standard structure during `create.project()` in order to tailor the environment to their own needs.

If the first thing you do after `create.project()` is to add a `.gitignore`, twiddle with the library settings in `config/globals.dcf` or add a knitr template with your favourite report layout, then this feature will be for you.  Perhaps your organisation or department has a standard set of analysis flow, reports or directory structure that you need to adhere to.  They can be set up once, and everyone can use them consistently across projects.

First of all, you set up a directory on your local system containing just the things you want to add to a standard `create.project()` layout (maybe a `README.md` replacement, perhaps a new `munge/01-A.R`).  Whatever you put in this folder will be merged into newly minted projects.  You can choose to overwrite `ProjectTemplate` files, or add new ones alongside.

Then you simply do:
```
templates("add", "<template name>", "local::/path/to/custom/template/dir")
```
Each time you now issue the command `create.project("<project-name>", "<template-name>")`  a new `ProjectTemplate` project is created that also incorporates your template automatically.

You can have as many templates as you want.  just type `templates()` to see what's currently installed.  If you have a particular template that you use a lot, you can set it to be the default:
```
templates("setdefault", <template-number-or-name>)
```
after which, every time you do `create.project("new-proj")` that template will be automatically applied, saving you some typing.

There is also the ability to use a template that someone has shared on github - maybe a scripted workflow for a particular problem domain.  To do this, you first add it to your template collection:
```
templates("add", "tm", "github:connectedblue/templates:tm")
```
This is a template I have created to assist with text data mining and analysis.  If you now do `create.project("my_tm_project", "tm")` then you will have a new project structure specifically tailored to get you started with text mining.  If you do this yourself as an example, you'll find a README which explains how to use this particular template to help analyse a big corpus of text.

By including the github option, I hope to enable the `ProjectTemplate` community to easily share analysis work flows among each other.  It also allows the functionality of `ProjectTemplate` to be extended and customised for lots of different situations without needing to have the core code updated.

If you would like to test drive this feature while this PR is under review, you can get the functionality by
```
library(devtools)
install_github("connectedblue/ProjectTemplate2@custom")
```
There's a lot more beside what I've described here.  Type `?templates` to get the full feature set, including special handling for `global.dcf` templates.

I look forward to seeing comments and feedback.


connectedblue:
Hmmm ... travis seems to be playing up at the moment.  The build just stopped during the last commit while it was installing packages.  The previous commit checked out OK, and there should be no reason this shouldn't either.

This has happened to me before - I'll retry again tomorrow.

connectedblue:
OK, this is puzzling.  I reverted back to a commit which was working earlier and I get the same travis error message:

```
ERROR: configuration failed for package ‘RODBC’
* removing ‘/home/travis/R/Library/RODBC’
Error: Command failed (1)
Execution halted
The command "Rscript -e 'deps <- devtools::dev_package_deps(dependencies = NA);devtools::install_deps(dependencies = TRUE);if (!all(deps$package %in% installed.packages())) { message("missing: ", paste(setdiff(deps$package, installed.packages()), collapse=", ")); q(status = 1, save = "no")}'" failed and exited with 1 during .
Your build has been stopped.
```
I'm not really sure what's going on this with this.

Issue #153, opened 2016-09-12
Ignore files in data folder


Hugovdberg:
How do you guys feel about adding an ignore list for file in `data/` to the configuration? A specific problem that occurs to me rather frequently is that at my workplace we're using windows, which creates Thumbs.db files as soon as a thumbnail is created.
Unfortunately, depending on the data files used, this is also created in the `data/` folder, after which the `load.project` treats it as a SQLite database, which it is not, and so it raises an error. On top of that, windows is notorious for keeping files locked, especially those thumbnail databases, and so deleting that file is often not an option. I can prevent windows from creating those files for myself, but I cannot force all my collegues to change their settings, so it is something I have to live with.

Now I see two possible solutions:
1. A hardcoded ignore of any `Thumbs.db`
2. An ignore list in the configuration that lets people ignore the files they want.

I think the latter is the most elegant, as for some people `Thumbs.db` could possibly be the file they want to import, and it also could speed up data loading if not all data is necessary in all stages of the project.

KentonWhite:
Since I don't work on windows, I've never encountered this problem -- which is why open source is so awesome! Such a common problem for so many users and such a simple solution.

@Hugovdberg do you think you could put together a simple PR with the ignore list and adding `Thumbs.db` to the default list?

Hugovdberg:
I can definitely create the code to do this, I was just wondering if the team would be interested to incorporate the code versus keeping it lean and clean. Are there other files or directories that might be created by an operating system automatically which we wish to exclude by default? I know I've seen some `.DS_Store`/`.Trash` files from Mac OS but I'm not sure when and where these are created so if they ever show up in a `data` directory.

KentonWhite:
Files like `.DS_Store` and `.Trash` are all hidden files. Is the `Thumbs.db` also a hidden file on Windows (don't know enough about the OS). If `Thumbs.db` is also a hidden file, might it an alternative solution be to ignore hidden files?

Hugovdberg:
`Thumbs.db` is hidden indeed, so it's a viable option to ignore hidden files, if it is possible to reliably detect hidden files across all operating systems. I will see what I can do on that. It would better fit the convention over configuration philosophy of the project, even though configuration in this case could drastically improve performance of analyses of subsets of the project data.

connectedblue:
My suggestion would be a `dataignore` file in the data directory which has a list of regular expressions to ignore when reading the data.

[...]

Issue #165, opened 2016-11-16
use of .R files in the data directory


connectedblue:
I've run into an issue on one of my projects which @KentonWhite highlighted when #160 was being developed to auto cache data during load.project()

This file demonstrates the issues, and how I worked around them:

https://github.com/connectedblue/capstone/blob/master/capstone/data/corpus.R

In this project, I am trying to create a test data set from the original data during my exploratory analysis (I use a custom config variable flag to toggle between the full and testing set to save rewriting the data acquisition code).

The way `.R` files are treated in the `data` directory results in the following issues:

- multiple variables can be created which may or may not be intended for project analysis.  In my file, the variables `corpus_dir`, `docs` and `doc` are side effects used to create the intended variable `corpus`.  I wrapped all the processing into a function so that these variables did not find their way into the global environment.
 - unless a variable is called the same name as the `.R` file, the data file will always be processed in the `data` directory,  This duplicates processing unnecessarily when the variables are already in memory or the cache.  I got around this by making the generated variable the same name as the file, and it is skipped during `load.project()` as a result.

These same issues would also occur in other data loaders such as excel and sql where the datafile can generate multiple global environment variables.

I have a proposed solution for the `.R` problems above, but I want to get the community's view before submitting a PR.

Essentially I propose using convention over configuration, and enforce a rule that `.R` files in the `data` directory can only produce one global variable with the same name as the file.  Any other global variables would be deleted which means that side effect variables like mine above don't have to be wrapped in a function to avoid polluting the global environment.

We'd have to put some migration logic in to warn the user about any legacy code that would need to be looked at.

I think it's good practice anyway to have one variable per `.R` file in the `data` directory, and enforcing it brings more benefits to users when they run `load.project()` making it faster to load by relying on memory and disk caching.

The change would be quite straightforward - I'm happy to do it, but I don't want to get distracted from my project so it'll be a couple of weeks before I raise a PR.  In the meantime, I'd like to get feedback on the approach to this issue.

KentonWhite:
@connectedblue I'm not quite following the problem. The convention of ProjectTemplate is that only variable names that match the file name will be checked prior to processing the file in the `data` directory.

In my opinion keeping the global namespace clean is the responsibility of the user. If the user wishes to create multiple global variables during the data step this is their choice. Sometimes the intended behavior is to create multiple global variables, as is the case when an excel workbook with multiple sheets is loaded.

[...]

PR #178, opened 2016-12-30
Ignore files in data directory

Hugovdberg:
This is an implementation of the feature discussed in issue #153. It adds a function that parses the `data_ignore` option in `global.dcf` and creates a regular expression that is evaluated for all files in the `data` directory. Some basic information about the functionality is on the configuring page of the website, `mastering.markdown` explains the feature in more detail.
Besides the changes in content a lot of trailing whitespace was removed from the files I edited by RStudio, sorry for the long list of deletions.


Hugovdberg:
Sorry, apparently I forgot a reload. The merge of my changes to `load.project` with the upstream changes isn't complete. Will fix it now.

connectedblue:
It's always nice to get a green tick :)

It would be relatively straightforward to add a test into `test-config.R` (you could write a new one based on those already there).  I would suggest adding `*.csv` to the `config$data_ignore` parameter, copy in `inst/example_data/example_01.csv` into the `data` directory, run a `load.project()` and test that the variable `example_01` doesn't exist. 

For belt and braces, you could then reset `config$data_ignore` back to `NULL`, run a `load.project()` and test that the variable `example_01` now exists.  

(BTW - what does actually happen if someone leaves the `data_ignore` parameter blank? )

Hugovdberg:
An empty parameter is considered a literal filename as it doesn't start and end with a `/`. Literal filenames are always matched from start to end by prepending `^` and appending `$`, resulting in the regular expression `^$`. So a blank option will actually match only a zero-length filename directly under `data/`, which can never exist as far as I know. Therefore leaving the parameter blank will match none of the files under `data/` which is exactly as expected.

I don't have any experience with `testthat`, but I'll see how far I can get!

connectedblue:
You'll find it quite therapeutic I think - always good to know that future developments won't unwittingly break your change.  Let me know if you have any questions.

I would have found this feature very useful in my last project when I had a lot of corpus data floating around `data` that I wanted to ignore most of the time. 

I'll install your branch this evening and test it out a bit.

Hugovdberg:
[offtopic]Sorry I wasn't able to push this earlier, first I was involved in the same capstone project as you were, and later the dayjob needed my full attention. I switched to the next session starting February as I just couldn't find the time to get my model to produce any sensible results..[/offtopic]

connectedblue:
Yes, same here... It has been such a fabulous course that I didn't want to rush a piece of rubbish for the capstone. So I also delayed to give myself more time. 

Perhaps we should collaborate on ideas to get a better model.  If we take a look at each others repo, we might get some ideas ...

connectedblue:
Hi @Hugovdberg 

I've installed your branch but am having difficulty in getting it to work.  I have a file named fix_locations.csv in the `data` directory and I set the following in `globals.dcf`:

```
data_ignore: /*\\.csv/
```

I tried various other combinations but couldn't get it to work.  What's the format I should use to ignore `.csv` files?

Also, are the slashes around the regular expression really necessary?  Couldn't each expression be treated as a regexp, so I would write `Thumbs\\.db` for example?

Hugovdberg:
There is a note in the documentation, you shouldn't escape your backslashes in the `global.dcf` unless you mean a literal `\`. The pattern should thus be `/.*\.csv/`, or simply `*.csv`, depending on whether you want to exclude all csv files including those in subdirectories (the former), or just the csv's directly under `data/` (the latter).

I made the decision to treat the strings as literals by default to make adding a single file as simple as possible. As written in the documentation, wildcards are allowed for both filenames and directory names. Regular expressions are quite advanced to my opinion and they require some more in depth knowledge to make sure you exclude exactly the files you want to exclude (as you just discovered yourself ;-) ).

Hugovdberg:
Actually, during writing tests I realised `*.csv` will ignore all csv files, including those in subdirectories, since the wildcard also matches the path separator. I think this behaviour is actually to be expected, and should be updated in the documentation.

connectedblue:
@Hugovdberg  yes, it worked for me.  It's nice and clean to just specify `*.csv`.

I realised that running it several times, the data was being loaded from the cache - i switched off `cache_loading` and `cache_loaded_data` and it all worked repeatably well.  I didn't look at the test cases you just posted, but I guess you reached the same conclusion.

connectedblue:
Nice set of tests @Hugovdberg :). 

One more thing to keep up to date:  The man page for `?project.config` needs updating so that the documentation is available right there in `R`.  You can use the same text as for the website, but the formatting is a little different (you'll see when you look at `project.config.R`).  You'll have to rebuild before committing and the man pages get generated automatically.

There's no need to include information about the default settings in `project.config` because those values are picked up auto-magically from the package files.

tests/testthat/test-load.R
```diff
@@ -28,40 +28,42 @@ test_that('user commands fail when not in ProjectTemplate directory', {
...
         on.exit(unlink(test_project, recursive = TRUE), add = TRUE)
-        
+
         oldwd <- setwd(test_project)
         on.exit(setwd(oldwd), add = TRUE)
-        
-        
+
+        # clear the global environment
+        rm(list=ls(envir = .TargetEnv), envir = .TargetEnv)
```
connectedblue: You could also use the `clear()` command here (also guarantees to remove any global variables that happen to begin with a dot.

Hugovdberg: I copied some code from the other tests since that test started failing because it assumed the environment was clear instead of making sure it was clear by itself. `clear` is better indeed, I will review the other tests as well.

Hugovdberg:
I noticed that cached data was loaded as well while writing the tests. I "fixed" the test for now by disabling caching during the test, but I think we can do better than that by building the list of ignored files and the corresponding cachefiles from those files.
The only problem I can see is in R files which potentially generate other variables than the filename implies, I'm not sure if and how these are written to cache. Then again I don't expect people to create huge datasets from an R file, and if the R file isn't an R file that can be loaded (just as the Thumbs.db can't be loaded by R) the ignore pattern will be added before it is cached. Those edge cases should be documented though.

Regarding `project.config`, I believe that file was created after I started developing the feature originally. I missed that while merging my changes into the most recent upstream master as there were no conflicts there. I will add the feature documentation there as well.

Hugovdberg:
I'm thinking `.load.cache` and `.load.data` should be integrated. That way it is very easy to get the list of variables to be imported, check whether it's cached and if so load from `cache`, otherwise load from `data`. We should also add a function `list.data` that checks all files in `data/`, tests if they should be ignored, whether they are cached and which reader will be used for the file. It returns a data.frame that can be used directly by `load.data`, but which can also be used by the user to see if the correct files are ignored.
If you guys agree we should put this PR on hold to refactor the loading process.

connectedblue:
@Hugovdberg  I think that your test is fine - disabling caching was important to see the effect of the feature during the test.  In real life, you don't need to combine ignore files and cache files because if they are ignored, they won't be in the cache by definition (if someone adds the `data_ignore` flag after a file has already been cached, all they need to do is a `clear.cache(var)` ).

Your point about R files is a good one (I raised a similar issue in #165).  I think there is a separate solution to that problem which doesn't interfere with this PR.

I would suggest keeping this PR as it is, and looking at any potential refactor as a separate PR - otherwise too many things will be packed into one change and it'll be hard to see the wood for the trees.

Hugovdberg:
Ok, I'll write a note in the documentation about ignoring already cached files, and create a new PR for the refactored `load.project`.

connectedblue:
Hi @Hugovdberg I think you're trying to do too many things at once in this PR.  There's no need to amend other tests here, especially if they are already passing.  The comment about using `clear()` was really for your new test.  

It's probably not a good idea to update unrelated tests otherwise it might seem as though a regression test is being rewritten to fit the new feature (which I know is not the case here)

connectedblue:
I think the last couple of commits have got the purpose of the PR tangled. 

The tests at `5a49d20` were running just fine - there was no need to adjust the other tests.  As a point of principle, PR's should not adjust previous tests without good reason (e.g. if an existing test needs to change because new functionality makes old behaviour obsolete).  Assuming that existing tests aren't working correctly is a potential slippery slope ... 

I'd be happier if you rolled back to `5a49d20` and re-applied just the documentation changes.   Then you can see for sure that your PR has not broken anything, and has added a new passing test.

Hugovdberg:
I'm not sure I understand what you mean. Don't you mean to say I should roll back to `277f952`, right before `5a49d20`? Because after that I only changed the documentation and made the test for the error on `Thumbs.db` more general as Travis tripped over the fact that the error message did not contain the same words as on my machine apparently.

connectedblue:
Oh yes, sorry - I meant `277f952` - the one that had the first working set of tests from Travis.  I've got a bit confused with the subsequent issues - I think you're saying that there are further test issues to resolve as well as the documentation?  

I think all I'm saying is to keep the commits in this PR focused just on the `data_ignore` feature.  It may well be that some other tests could be coded more efficiently - but as long as they work and are unrelated to `data_ignore`, then it's best to leave them alone.
`

Hugovdberg:
I reverted the changes to `test-load.R`, keeping all subsequent edits to other files. This produces again a long list of edits, but before the final merge into `master` I will squash all commits into one. Or should I do that now to make the review easier?

connectedblue:
I think you can keep them as they are in the PR - @KentonWhite usually squashes them into one commit when he applies to master.

Sorry for confusing earlier - I see that I applied by comment about `clear()` against the wrong line.  Anyway, all seems good now.

Hugovdberg:
Let's hope @KentonWhite also thinks it's ok, I'm finishing up the documentation to a new `list.data` function which already assumes there is such a thing as ignored data :-)

R/load.project.R
```diff
@@ -156,9 +157,44 @@ load.project <- function(override.config = NULL)
...
+  paste0(c(literals, regexes), collapse = '|')
+}
+
+.load.data <- function(recursive, ignore_files = NULL) {
   .provide.directory('data')
   data.files <- dir('data', recursive = recursive)
+  ignore.index <- grepl(.prepare.data.ignore.regex(ignore_files),
+                        data.files,
+                        perl = TRUE)
+  # print(data.frame(files = data.files,
```
connectedblue: Probably worth removing this debug statement

R/project.config.R
```diff
@@ -4,62 +4,66 @@
...
 #' @details The options that can be configured in the \code{config/global.dcf} are
 #'  shown below
 #' \tabular{ll}{
 #'  \code{data_loading} \tab This can be set to 'on' or 'off'. If data_loading is on,
-#'  the system will load data from both the cache and data directories with 
+#'  the system will load data from both the cache and data directories with
 #'  cache taking precedence in the case of name conflict. \cr
+#'  \code{data_ignore} \tab A comma separated list of files to be ignored when importing
+#'  from the \code{data/} directory. Regular expressions can be used but should be delimited
+#'   (on both sides) by \code{/}. Note that filenames and filepaths should *never* begin with
```
connectedblue: I'm not sure this is markdown - the asterix would just get carried through to the man page I think.  

I don't think emphasis is used in man pages.

R/load.project.R
Hugovdberg: whoops, I added it commented in the dreaded commit I rolled back, should have rolled back this edit as well. I actually built a new function `list.data` instead that should be called in the future so this debug statement becomes obsolete overall. That branch however is a linear descendant of this branch, so I cannot create a pull request or these edits are included twice..

R/project.config.R
Hugovdberg: Thanks, missed that one, fixed, but waiting for final comments before committing.

KentonWhite:
@Hugovdberg @connectedblue thanks for carrying on over the holidays. Sorry I've been traveling and not very active.

I like the PR as it is.

I'm waffling on the choice of defaults. On the one hand, Thumbs.db is windows specific -- does it make sense to have a default that only applies to a single operating system? What if users on other operating systems wish to have their system files included in the defaults? We start to get a long list of ignored files! Could it be cleaner if we detect the OS when the project is created and adjust the defaults accordingly? The trade off is that this becomes tricky to test.

On the other hand, including every possible system file for every possible OS in the ignore defaults does make projects more portable. We can be assured that people moving a project from Linux to Windows will have the same environment because Thumbs.db would not have been accidentally loaded. Also having all OS ignore files listed will keep the ignore list clean when people are collaborating on projects using different OS's.

As you can see there are good arguments on either side. Before merging I'd like to get your thoughts!

connectedblue:
Hi @KentonWhite, good to see new back, and Happy New Year.

My thought is this is exactly the sort of use case that custom templates is designed for in #176 (closed for now just to re-think the implementation).  So users on Windows who like git might have one set of needs which can be templated (and shared), whereas mac users who use Subversion might need something different.  The custom templates specifically allow pieces of `globals.dcf` to be in the template and merge in correctly to newly created projects.

So, I would say ship with no defaults and let custom templates get designed to handle different use cases (maybe there's even a github repository of them somewhere).  That way the `ProjectTemplate` core stays clean of implementation environments.

I realise this is a bit of a "jam tomorrow" answer as we don't have the custom templates yet, but hopefully it'll be soon.

connectedblue:
Another option might be to leverage the functionality of #169 and put code in as the default value - something like `data_ignore: 'local_ignore()'` and then a skeleton file `local_ignore()` is included in the libs directory which could do a local OS dependent read, and also allow users to add their own.  

If this solution appeals, it should be done in a separate PR though

Hugovdberg:
I don't think ignore options should be platform dependent. When people collaborate on a project they might not all use the same platform. Initially we discussed adding a pattern to ignore filenames starting with a dot, but hidden files are ignored already by `list.files`/`dir`. So if there are more files that should be ignored we should add them to the defaults.

connectedblue:
@Hugovdberg the `local_ignore()` approach doesn't have to return a hard coded list - it can inspect the OS type at `load.project` time and make some pre-canned OS-specific choices.  This guarantees portability.  It can also return some optional, additional user-specified ones.  If the whole function is provided as a single file in the  `lib` directory the user can delete it entirely if they want, or they can add to it.

I think it's potentially a cleaner approach, otherwise this line of `globals.dcf` becomes a cluttered mess of default hard coded options which users will find confusing to customise to suit their needs.  

However - this doesn't have to be done now.  It's an option in the future if the list of defaults gets longer and longer

Hugovdberg:
I agree settling on the defaults can be done in the future, Now I see what you mean by `local_ignore`, calling it in `load.project` could be a flexible and clean solution indeed. By adding it to the default settings it can be easily removed if deemed unnecessary by the user.

connectedblue:
Yes, one for the future I think.  

Just to be precise, `load.project` wouldn't need to know about, or call `local_ignore` directly.  If the config is coded as ``data_ignore:`source('lib/local_ignore.R')` `` then  when `.load.config()` is executed during `load.project` the `data_ignore` value will be dynamically calculated by whatever `local_ignore.R` returns.

KentonWhite:
So are we going to remove defaults for now and work on a `local_ignore` + custom templates for the future?

Hugovdberg:
I've removed the defaults for now, but I'm not so sure whether a `local_ignore` should be added. The trouble is we can't really build a function that works for everyone, and if we let the users build the list it is essentially part of their template. So I'd say make a great custom templates functionality, and let the users ignore what they want.

Issue #177, opened 2016-12-29
Change /doc to /docs


Hugovdberg:
Recently GitHub made publishing a website on GitHub Pages a lot easier: [see this blog post](https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3). There are two options to set the Pages root in the `master` branch, either use the root directory or the `docs` folder. By renaming the `doc` folder to `docs` in new projects we facilitate the latter option.

As the `doc` folder isn't included in the default workflow and mainly for convenience as far as I can see it should not have a major impact on our project.

connectedblue:
Nice - I hadn't seen that post.  I keep meaning to learn jekyll etc, but this looks super easy.

Agreed that your proposed change is relatively minor.  There would also need to be a change to the `exclude` variable in `create.project` so that minimal projects keep working correctly.

I would also suggest putting some detection code in `migrate.project` to change the directory name in existing projects, with a message to the user to explain that this has been done.

Hugovdberg:
I'm not sure about changing existing projects, existing links to the `doc` folder might break and cannot be changed automatically. I would suggest to create a message pointing to the blog post to notify the user of the possibility, or at the very least prompt the user whether the folder should be renamed.

connectedblue:
Yes, there are edge cases either way.  The purpose of `migrate.project` is to bring a previously created project up to date with the latest version (as if `create.project` has been run again).  

If the `doc` directory remains after migration, then `migrate.project` loses its integrity for all users just because some might have broken links.  

A user prompt would be a nice middle ground.

KentonWhite:
I agree this would be a nice change. Change existing projects through the `migrate.project` function. I would not erase or rename the `doc` directory with `migrate.project` but instead simply add the `docs` directory to existing projects. This prevents those projects from breaking. The user can then delete the `doc` directory if they desire. New projects should only have the `docs` directory.

connectedblue:
@KentonWhite neat solution.  Best of both worlds - forward and backward compatibility :)

Hugovdberg:
Simply creating the `docs` directory is indeed a better option, but I think there should be a message that the new `docs` isn't just a typo (for the people who missed the update). So tasks to do:

- [ ] Change skeleton `doc` to `docs`
- [ ] Add "create `docs`" to `migrate.project()`, including message when folder is created
- [ ] Update our documentation with reference to GitHub blog post

Missing anything?

[...]

Issue #179, opened 2016-12-30
Review creating new directory structure at project creation and migration


Hugovdberg:
Issue #177 proposes a new name for the `doc/` directory to `docs/` to enable easy integration with GitHub Pages. A change was suggested to add renaming the folder to `migrate.project()` as well as the purpose of that function is to get the project up to the same level as after a call to `create.project()`. However, at this moment new directories added to the skeleton structure aren't created at all during migration.
This is partially because it is possible to call `create.project` with the `minimal = TRUE` option, which does not create all directories from the skeleton. This option should be honoured during project migration, but at the moment it is hard to reliably detect what setting was used. It would be best I think to store the setting during project creation, and be able to overrule it in `migrate.project` so an existing minimal project can be upgraded to a full project.

Also, `merge.strategy = allow.non.conflict` could perhaps be made more lenient because even if only an empty directory exists with a conflicting name (which would not mean any data loss if overwritten) the `create.project()` functions raises an error. I think it would be nice if it would only fail on overwriting files.

Combining all these remarks I think we should refactor `create.project` and `migrate.project` to call the same functions in the background, exposing a similar signature. Consider the following pseudo code:

```r
create.project <- function(project.name = '.', minimal = FALSE, dump = FALSE,
                           merge.strategy = c("require.empty", "allow.non.conflict")) {
  if (!dir.exists(project.name)) {
    dir.create(project.name)
    setwd(project.name)
    .update.project.structure(minimal, dump, is.update = FALSE)
  } else {
    setwd(project.name)
    if (allow.non.conflict) {
      compare.existing.files
    }
    .update.project.structure(minimal, dump, is.update = FALSE)
  }
}

migrate.project <- function(minimal = .is.minimal.project()) {
  .update.project.structure(minimal)
}

.update.project.structure <- function(minimal, dump, is.update = TRUE) {
  if (minimal) {
    exclude dirs from skeleton
  }  
  for (dir in skeleton) {
    if (not exists dir) {
      create dir
      if (is.update) {
        message dir created
      }
    }
  }
  for (file in skeleton) {
    if (not exists file) {
      copy file
      if (is.update) {
        message file created
      }
    }
  }
  if (dump) {
    dump files
  }
  for (setting in skeleton) {
    if (not exists setting) {
      add setting
      if (is.update) {
        message setting created
      }
    }
  }
  if (is.update) {
    for (setting in settings) {
      if (setting not in skeleton) {
        delete setting
[...]

PR #180, opened 2016-12-30
Renamed doc to docs and added references to GitHub Pages configuration

Hugovdberg:
Implementation of issue #177, changed the skeleton, and added relevant references to GitHub Pages (the rationale behind the change). A message is issued when `migrate.project` is called in a project where the `doc` directory exists and `docs` does not, after which the new `docs` directory is copied from the skeleton.

### DIFF
--- a/R/migrate.project.R
+++ b/R/migrate.project.R
@@ -26,7 +26,8 @@ migrate.project <- function()
 
   # Initialise migration flags
   config_conflicts <- FALSE
-  directory_conflicts <- FALSE
+  version_conflict <- FALSE
+  other_conflicts <- FALSE
   config_warnings <- NULL
 
   # Flags stored in environment env
@@ -45,15 +46,14 @@ migrate.project <- function()
       suppressWarnings(.load.config())
     }
   )
+  version_conflict <- .check.version(loaded.config, warn.migrate = FALSE) != 0
 
   # Detect other migration issues
   doc_not_renamed <- dir.exists("doc") && !dir.exists("docs")
-  directory_conflicts <- any(doc_not_renamed)
+  other_conflicts <- any(doc_not_renamed)
 
   # Exit if everything up to date
-  if ((.check.version(loaded.config, warn.migrate = FALSE) == 0) &&
-      !config_conflicts && !directory_conflicts
-    ) {
+  if (!any(version_conflict, config_conflicts, other_conflicts)) {
     message("Already up to date.")
     return(invisible(NULL))
   }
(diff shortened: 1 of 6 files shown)


R/migrate.project.R
```diff
@@ -17,118 +17,127 @@
...
+
+      assign("config_conflicts", TRUE, envir = env)
+      assign("config_warnings", w$message, envir = env)
+      suppressWarnings(.load.config())
+    }
+  )
+
+  # Detect other migration issues
+  doc_not_renamed <- dir.exists("doc") && !dir.exists("docs")
+  directory_conflicts <- any(doc_not_renamed)
```
connectedblue: I would name this `docs_directory_conflict`  just be be super specific that this migration check is only about the `docs` directory and none of the others.

R/migrate.project.R
```diff
@@ -17,118 +17,127 @@
...
-  
-  
-  
+  if (doc_not_renamed) {
+    message(paste(
+      "",
+      "The doc directory has been renamed to docs for new projects. This was",
+      "done to make it easy use the new option for GitHub Pages to use the docs",
+      "directory as the root for your website.",
+      "The new docs directory has been created in this project, but no files have",
```
connectedblue: I wouldn't mention links breaking here because there are no `ProjectTemplate` workflow that uses `doc`.  If the user has designed their own workflow around `doc` then this could be all sorts of things, not just links.  Also, the first sentence sounds as if `migrate.project` has renamed the directory.

Perhaps the message should be something like 
```
The doc directory has now been deprecated and ProjectTemplate now uses docs instead. This is to ease integration with GitHub Pages (See https://github.com/blog/2289-publishing-with-github-pages-now-as-easy-as-1-2-3).

A new docs directory has been created in this project, but your existing doc directory remains.  Please review and move manually if you wish.
```

Hugovdberg: Great suggestion, I struggled to get a good message. You've got to love PR's for this ability to get feedback on specific edits!

R/migrate.project.R
Hugovdberg: Wel this variable naming was more or less based on the discussion in #177 and #179. The comments in `migrate.project` already provided the general structure, and I thought it better to provide a means for easy expansion.
Perhaps I should rename `directory_conflicts` to `other_conflicts`, that way any other conflicts can be added easily without the need to add every conflict to the if statement on line 54.

connectedblue: I was only meaning that `docs_directory_conflicts` is more specific than `directory_conflicts` so it might be more descriptive (ie, it's not a conflict to do with say the `reports` directory).

I prefer the way you have coded the if statement because each conflict needs a specific migration action, so having a more general flag wouldn't help.

R/migrate.project.R
```diff
@@ -17,118 +17,127 @@
...
+        collapse = "\n"))
+    }
   }
-  
+
   # Process other migration conflicts
-  
-  
-  
+  if (doc_not_renamed) {
```
connectedblue: I'm confused here - shouldn't it be `directory_conflict`?

In fact, aren't these two always going to be the same?  One looks redundant.

Hugovdberg: At this moment `doc_not_renamed == other_conflicts` (see line 53 in ba584aa). This allows for easy expansion later, simply add another flag for the specific case. `other_conflicts` like the `config_conflicts` just says there is some conflict, and after the `if` statement on line 56 the specific conflicts are tested for and solved as good as we can.

R/migrate.project.R
```diff
@@ -45,15 +46,14 @@ migrate.project <- function()
...
   # Detect other migration issues
   doc_not_renamed <- dir.exists("doc") && !dir.exists("docs")
-  directory_conflicts <- any(doc_not_renamed)
+  other_conflicts <- any(doc_not_renamed)
 
   # Exit if everything up to date
-  if ((.check.version(loaded.config, warn.migrate = FALSE) == 0) &&
-      !config_conflicts && !directory_conflicts
-    ) {
+  if (!any(version_conflict, config_conflicts, other_conflicts)) {
```
connectedblue: nice

PR #181, opened 2017-01-01
Prune unused code

connectedblue:
This PR proposes a small cleanup to two functions:

- Remove unused function `.get.template.tar.path()` in `create.project`.  This doesn't get called from anywhere and doesn't appear to be relevant for how templates are created.  Perhaps it is a relic of how skeleton projects used to be created?  
 - Remove two commented out code chunks in `load.project`.  Again, looks like a design decision taken in the past and not relevant now.

No user functionality changed as a result of this PR.

### DIFF
--- a/R/create.project.R
+++ b/R/create.project.R
@@ -133,9 +133,6 @@ create.project <- function(project.name = 'new-project', minimal = FALSE,
   )
 }
 
-.get.template.tar.path <- function(template.name)
-  system.file(file.path('defaults', paste0(template.name, ".tar")), package = 'ProjectTemplate')
-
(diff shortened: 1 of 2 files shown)


Hugovdberg:
The `.get.template.tar.path` function is much like a basic version of the `.get.template` function I suggested in #179 ;-) but I don't see any problems in removing those lines of code.

connectedblue:
Yes, that was what prompted me to raise this PR - my brain hurts when I try to link all these things in my head.

Hopefully this PR will reduce the cognitive burden for `ProjectTemplate` developers just a little :)

PR #183, opened 2017-01-02
Make reload.project() compatible with clear() and clear.cache()

connectedblue:
This PR proposes an extension to the `reload.project()` function to take account of recent additions to `ProjectTemplate` in a more natural way.

The existing behaviour is to clear the global environment and then run `load.project()` again. The PR essentially does the same but takes account of more factors:

- There is a new parameter `config$sticky_variables` which are for specific datasets that may take a long time to load and the user doesn't want them cleared during a `clear()`.  These are now also honoured during `reload.project()`

- There is a new paramer `reset` so if `reload.project(reset=TRUE)` is called, then everything is cleared from both Global env and  cache regardless (the project is reset).

 - The parameter `...` is used to capture any additonal parameters to pass to `load.project` (e.g. `override.config`).  This was hard coded before, but it's more generic to use the dots notation.

The documentation for `reload.project()` has been updated.  I couldn't see any relevant parts of the website documentation that should be updated.

### DIFF
--- a/R/clear.cache.R
+++ b/R/clear.cache.R
@@ -64,9 +64,3 @@ clear.cache <- function (...){
                 message("Removed successfully from cache\n")
         }
 }
-
-# Determine if a variable is cached or not
-.is.cached <- function (variable) {
-        file.exists(file.path(.cache.dir, paste0(variable, ".RData")))
-}
-
--- a/R/clear.cache.R
+++ b/R/clear.cache.R
@@ -65,3 +65,8 @@ clear.cache <- function (...){
         }
 }
 
+# Determine if a variable is cached or not
+.is.cached <- function (variable) {
+        file.exists(file.path(.cache.dir, paste0(variable, ".RData")))
+}
+
--- a/R/clear.cache.R
+++ b/R/clear.cache.R
@@ -64,3 +64,4 @@ clear.cache <- function (...){
                 message("Removed successfully from cache\n")
         }
 }
+
--- a/tests/testthat/test-reload.R
+++ b/tests/testthat/test-reload.R
@@ -1,0 +1,155 @@
+context('Reload project')
+
+test_that('override.config is passed through correctly to load.project', {
+        test_project <- tempfile('test_project')
+        suppressMessages(create.project(test_project, minimal = FALSE))
+        on.exit(unlink(test_project, recursive = TRUE), add = TRUE)
+        
+        oldwd <- setwd(test_project)
+        on.exit(setwd(oldwd), add = TRUE)
+        
+        clear(force=TRUE)
+        
+        # Create some R code and put in data directory 
+        CODE <- paste0(deparse(substitute({
+                x <- 10
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/x.R")
+        
+        CODE <- paste0(deparse(substitute({
+                y <- 20
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/y.R")
+        
+        CODE <- paste0(deparse(substitute({
+                z <- 10
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/z.R")
+        
+        # reload the project but switch off data loading
+        suppressMessages(reload.project(override.config=list(data_loading=FALSE)))
+        
+        # x should not exist in Global Env or the cache
+        expect_true(!exists("x"))
+        expect_true(!.is.cached("x"))
+        
+})
+
+test_that('reload.project ignores sticky_variables', {
+        test_project <- tempfile('test_project')
+        suppressMessages(create.project(test_project, minimal = FALSE))
+        on.exit(unlink(test_project, recursive = TRUE), add = TRUE)
+        
+        oldwd <- setwd(test_project)
+        on.exit(setwd(oldwd), add = TRUE)
+        
+        clear(force=TRUE)
+        
+        # Create some R code and put in data directory 
+        CODE <- paste0(deparse(substitute({
+                x <- 10
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/x.R")
+        
+        CODE <- paste0(deparse(substitute({
+                y <- 20
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/y.R")
+        
+        CODE <- paste0(deparse(substitute({
+                z <- 10
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/z.R")
+        
+        # Read the config data and set sticky_variables to x,y
+        config <- translate.dcf("config/global.dcf")
+        expect_error(config$sticky_variables <- "x,y", NA)
+        write.dcf(config, "config/global.dcf" )
+        
+        # Load the project and check that x exists and is in the cache
+        suppressMessages(load.project())
+        expect_true(exists("x"))
+        expect_true(.is.cached("x"))
+        
+        # reload the project and config, x, y should not be cleared
+        expect_message(reload.project(), "not cleared: config x y")
+        
+        # Also, z should be reloaded from data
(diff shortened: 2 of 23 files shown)


R/clear.cache.R
```diff
@@ -65,3 +65,8 @@ clear.cache <- function (...){
         }
 }
 
+# Determine if a variable is cached or not
+.is.cached <- function (variable) {
+        file.exists(file.path(.cache.dir, paste0(variable, ".RData")))
```
Hugovdberg: Is this guaranteed to test the same condition as `.read.cache.info` does it, and do you think that the `in.cache` test in `.read.cache.info` should call this function instead of implementing two ways to check if the variable is in cache? They should do exactly the same and there's no better guarantee than coding it in one place only.
Also, I'm thinking this function should really be in `cache.R`, or all cache related auxiliary functions should be placed together in a single file.

connectedblue: Good spot @Hugovdberg.  You're quite right.  

When I was coding the caching I had an `.is.cached` function, but it became apparent that you need to know a lot more than the presence of the file (e.g. is it stale or not).  So `.is.cached` morphed into `read.cache.info`.

I'll call the function directly even though it does more work than I need here.  It's simpler than creating a new function and refactoring.  The code would read something like `if (.read.cache.info(var)$in.cache) ...` which is reasonably readable.

Hugovdberg: If you just want to return the `in.cache` status you could replace the body of the function with `.read.cache.info(var)$in.cache` without any if statements. But I think it would make more sense to do it the other way around and make `.read.cache.info` call `.is.cached`. That way `.is.cached` contains exactly the code to determine whether a variable is cached, and `.read.cache.info` can return the additional info necessary info for cache invalidation. There is no need for a new function, is it?

connectedblue: Yes, agreed.  There is no need for a new function.  I pushed the changes to the branch and the PR has been simplified.

Hugovdberg: Ah now I see why you mentioned the `if` statement. I think it would be nice to either make `.read.cache.info` able to check more than one variable at a time, or restore the `.is.cached` function. I want to read cached state as well in an upcoming PR (to add a `list.data` function), and I need to loop over all variables just to get the cached state. `file.exists` works fine with a character vector, but the part where the `cache.hash` is loaded doesn't work for more than one variable. I think `.is.cached` is useful standalone without loading the hashes, as well as being more descriptive than `.read.cache.info(variable)$in.cache`, and so I'd prefer `.read.cache.info` calling `.is.cached`. I'm sorry that I didn't express the point about a wider use of `.is.cached` more clearly before.

connectedblue: @Hugovdberg I've had a look again, and there are deeper issues to consider with an `is.cached` function.  For example:
 - should the existence of both `.RData` and `.hash` be checked, and only declared `TRUE` is both are present?
 - If `.is.cached` detects one missing, should it delete the other and report `FALSE` to maintain consistency?
 - should `is.cached` be called by the cache loading part of `load.project`?

Also, the requirements of other functions such as the one you are working on need to be thought through more fully as to what makes sense.

I think we should have this debate in an issue and then raise an appropriate PR.  I'll raise one now.

So, for this PR, I propose to stick with the existing `.read.cache.info()` function and update later when the above questions are resolved.

KentonWhite:
I agree that the expected behavior is that *all* data is removed and reloaded from scratch. I'm a bit worried about removing cache items using `reload.project()`.  

Let's imagine the user with all of their long loading data stored in cache. After an update they run `reload.project()` expecting the old behavior and for whatever reason some of those variables are missing from the global environment. `reload.project()` now wipes out their cache and they are stuck with a long rebuild time.

If we are removing items from cache with `reload.project()` I'd feel more comfortable if we provide a confirmation for cached items that will be removed. This way a user can prevent a disastrous mistake!

connectedblue:
I see your point.   

Can we break down the cases between how it might work going forward, and what a user expects after updating to the latest `ProjectTemplate`?  We can help the latter through `migrate.project` if we detect they have things in the cache.

So ..... should the new behaviour be:

 - if I `reload.project` I want everything re-calculated from scratch, so clear both memory and cache
 - if I have declared certain variables to be protected in `config$sticky_variables`, then I want them to remain in memory and/or cache during a reload
 - Even if I have declared `config$sticky_variables`, I want to consciously erase everything anyway, so I set `reload.project(reset=TRUE)`

I think points 2 and 3 are not contentious - the user is in control of what's happening.

With point 1, it boils down to whether most of the time users expect to have both `clear()` and `clear.cache()` when doing a reload.  

I did toy with the idea of creating a `reset.project()` function which did both and leave `reload.project()` to just clearglobal env - but this seems even more confusing for a user to have to remember.

On balance, I'm agnostic on clearing the cache in case 1.  I can see the benefits, and I can see the other side.    Let me rework the PR to just use `clear()`.  We can always add `clear.cache()` in later if required.

Are you still comfortable with case 3 - having a `reset` parameter explicitly set to clear everything?

connectedblue:
OK, this is the reworked PR for review.

The behaviour is essentially the same as before, except that `config$sticky_variables` is honoured and there is a new `reset` parameter to allow a complete scrub out and reload if that's what the user wants.

tests/testthat/test-reload.R
```diff
@@ -0,0 +1,155 @@
...
+        CODE <- paste0(deparse(substitute({
+                z <- 10
+                
+        })), collapse ="\n")
+        
+        # save R code in the data directory
+        writeLines(CODE, "data/z.R")
+        
+        # Read the config data and set sticky_variables to x,y
+        config <- translate.dcf("config/global.dcf")
```
Hugovdberg: why don't you use`.read.config` and `.save.config`? I think it's a little clearer and more portable if we ever decide to use another format to store the configuration (or move the location). See also #184, where some wild variations on this were replaced with `.read.config`.

connectedblue: In test scripts you want to test user functions, not internal ones.  My preference is to use functions in tests that are as close as possible to simulating what a user would do.  So loading a config file and resaving a particular variable should achieve a certain result.  A user would probably use an editor and save the file.  You cant really simulate that in a script, so I use `read.dcf` and `write.dcf` as the next best thing.  Sometimes I use other functions - it's no bad idea to have a variety of set up techniques in a test script.

The problem with using internal functions is they may contain new bugs that you would want to catch.  I'd rather a test failed during development so it could be investigated rather than have everything pass through cleanly.

Hugovdberg: I agree on testing user functions, but you're not testing the function (actually, that function isn't tested at all). I agree a wide spectrum of tests should be applied, but there is a distinction between the functions you use to setup the test environment (which should be tested separately and to which `translate.dcf` belongs in this case) and the functions you test in the test case. (Also, if you'd really want to simulate a texteditor why not use `readLines` and `writeLines`? ;-) )

KentonWhite:
I like the notation `reload.project(reset = TRUE)`.  It allows us to keep `reload.project` working as it has (reload the project, including reloading the cached data) and start moving towards a future where `reload.project` recalculates all cached variables. 

I would also add a test to see if the user explicitly passes `reset`. If they don't pass a `reset` variable, provide a warning describing the new feature and that in a later release the default will change from `FALSE` to `TRUE`.

Also curious why you closed this PR?

connectedblue:
Hi @KentonWhite.  I like this idea - a warning of future deprecation.  I'll have a go at that.  I'm not really sure how you test if a user actually specified a variable (other than checking for the default value), but I'll have a google around.

I closed the PR because I didn't want it merged until I've reviewed the tests.  I haven't got a lot of time now that the New Year is underway, so I'm not sure I'll get it done over the next few days.  I also want to add in your suggestion, so that'll be a bit longer again.

connectedblue:
Added the check suggested by @KentonWhite if the cache is not empty.

PR ready for review now.

connectedblue:
Hi @KentonWhite - have you had a chance to review this PR?  It was updated to reflect your last comments back in January.  

I've been rather busy at work since then, so haven't had a lot of time to look at `ProjectTemplate` recently.

KentonWhite:
Sorry hadn't seen that it had been re-opened

Issue #186, opened 2017-01-14
tests fail in master


connectedblue:
I just synced to the latest master and I have some tests fail:

```
Error in rsqlite_send_query(conn@ptr, statement) : 
  file is encrypted or is not a database
1. Failure: ignored data files are not loaded (@test-load.R#186) -----------------------------
exists("test", envir = .TargetEnv) isn't false.


2. Failure: ignored data files are not loaded (@test-load.R#193) -----------------------------
exists("test", envir = .TargetEnv) isn't false.

```
These are run locally, straight off the `master` branch with no modifications.

I'm not sure how this can happen as the travis checks must have picked this up at the last merge.  Does anyone else see this failure?

Hugovdberg:
It seems to me that `testthat` isn't always completely catching the errors, especially that `RSQLite` error. I also noticed that sometimes tests seem to succeed or fail depending on whether you run `devtools::test`, `testthat::test_package` or `devtools::check`. Did you experience differences between those? 
Perhaps this has something to do with the fact we're messing with the global environment during testing? Since (almost) every function we write uses `.TargetEnv` (I fixed a couple succeeding 'by accident' in an upcoming PR) we could perhaps use a clean `.TargetEnv` in the tests, with only some specific tests for the difference between `.TargetEnv == GlobalEnv` and `.TargetEnv == .TestEnv`?

Hugovdberg:
Also, `testthat::test_package('ProjectTemplate', 'create')` produces some failed tests on my system that are listed along the way but not in the final results. Does that happen with you to?

KentonWhite:
I'm seeing the this error:

```
Error in sqliteSendQuery(con, statement, bind.data) : 
  error in statement: file is encrypted or is not a database
```

But `testthat` is not showing it as a failure. Reading the comments this seems to be expected behavior:

```
 # Should trigger error
  # reload the project, now with an illegal Thumbs.db
```
So the illegal Thumbs.db is throwing and error, which is printed to the console -- but it is expected that there is an error.

connectedblue:
I just do `source(tests/run-all.R)`.

I didn't know there were multiple ways to run the tests.  Even if there are different environment issues, it still seems odd why just those tests fail, and moreover why Travis passed them.

Either way, my own development is at a standstill.  I don't want to add new features when tests fail, because you can't see whether it's because of the test, or because of a change I have made.

Hugovdberg:
[...]

PR #188, opened 2017-01-23
Remove warnings from Travis and devtools::test() checks

connectedblue:
This PR proposes some updates to the existing code base to prevent warnings thrown by some test tools.  This was discussed in #186.

The warnings fixed by this are:

- those thrown when `devtools::test()` is run in the `ProjectTemplate` directory
- when examining the travis console screen warnings were shown about the binding of the `config` variable in `.remove.sticky.variables()` function.

No user functionality is changed in this PR, it simply cleans up some underlying test cases to improve the regression suite.

### DIFF
--- a/R/clear.R
+++ b/R/clear.R
@@ -60,8 +60,10 @@ clear <- function (..., keep=c(), force=FALSE) {
 
         # If we're in a project template directory, load a copy of config to
         # make sure it's the latest into the global env  
-        if (.is.ProjectTemplate()) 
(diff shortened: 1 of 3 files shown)


connectedblue:
Hi @KentonWhite Could you review and merge this PR if you're happy with it?

It touches some files which I'll need to modify to address the bug in #192 so I'd rather develop off the master with the cleaned up tests.

KentonWhite:
Merged.  Thanks!

Issue #192, opened 2017-04-19
cache bug


eribul:
I created a project with older version of ProjectTemplate and it worked fine. I then updated the package and got the error message bellow running `cache` on an object that was already cached. The problem was solved by first removing the old cached object and then rerunning the `cache` function again (then a .hash file was created along with hthe cached file etc and everything seems to be in order).  

```{r}
Error in if (stored$hash["VAR", ]$hash == genv.hash["VAR", ]$hash) { : 
  argument is of length zero
```

eribul:
BTW: new version I updated to is 0.7

connectedblue:
Thanks for reporting @eribul - that is definitely a bug.  The new caching mechanism has new information in a `.hash` file as you discovered, but the user shouldn't be exposed to that.

I'll look into this in more detail.  I think the fix is to do some work in `migrate.project` to fix this automatically.

Can you tell me:  did you run the `migrate.project()` command after you upgraded to the latest version of `ProjectTemplate`?

connectedblue:
Fix submitted to address this issue #194

KentonWhite:
@eribul The fix has been merged into master.  I'm closing this issue and if the fix doesn't work for you please feel free to reopen!

PR #194, opened 2017-04-20
Bug Fix:  Missing hash file causes failure in subsequent cache() operation

connectedblue:
This big was raised by @eribul in #192.  It was reproducible exactly as described in the issue.  

The cause is because there should always be a `.hash` file present in the `cache` drectory alongside the `.RData` file.

The solution is as follows:  If the hash is missing, either because there was a previously cached variable made under an old version of `ProjectTemplate`, or it may have been inadvertently deleted, then the associated cache file is also removed when a `cache` command is issued again.  This ensures a successful recache and a consistent `cache` directory again.

A test set has been added to check for this use case.

Note that there are a large number of `man` files re-generated when I rebuilt the package for this submission.  These were not as a result of a change made by me under this PR.  I suspect that an upgrade to ROxygen to v6.0 may have forced a recompile.  Either way, the documentation is up to date now, so it's not a big deal.


KentonWhite:
Thanks!

Issue #170, opened 2016-12-03
csv2.reader() does not behave as expected with regard to read.csv2()


crubb:
Hey guys,

when using the .csv2 file extension I was expecting that R's `read.csv2()` would be used instead of `read.csv()`.

From R's documentation:

`read.csv(file, header = TRUE, sep = ",", dec = ".", ...)`
`read.csv2(file, header = TRUE, sep = ";", dec = ",", ...)`

ProjectTemplate's implementation when using `csv2.reader()` is using `;` as a separator, but keeping `.` as the decimal separator. Since the main reason for using `;` as a separator would be having `,` as a decimal delimiter (standard CSV file format e.g. in Germany), I would propose changing `csv2.reader()` to comply with R's `read.csv2`.

Since this might be a breaking change, I didn't put together a pull request, yet.

What do you think?

Best regards,
Christian

connectedblue:
@crubb that seems to make sense to me - although I personally don't use `.csv2` files.  You might break a few regression tests though which would have to be fixed :)

On a related point, I've wondered whether we should just be using `read_csv`  function from the `readr` package because that does a lot of auto checking on the various csv types and makes sensible decisions on data types etc.  I wonder if we did this, would it do away with the need to implement `csv2.reader` altogether...

Hugovdberg:
`readr` provides both a `read_csv` and `read_csv2` function, similar to the `read.csv` and `read.csv2`. There would still be a need for a `csv2.reader` because `read_csv` doesn't magically determine the column separator. I agree with @crubb that the current `csv2.reader` doesn't work consistently with the `read.csv2` function which the name implies. I have the same issues with the csv2 format (being Dutch, using the same decimal comma).

KentonWhite:
Looks like many people are in favour of moving toward `read.csv2`. Since this is a breaking change, I would for now include a warning in the `csv2.reader` detailing the change and advising users to check their csv data. We would remove this warning after sufficient time.

eribul:
Thumbs up from a Swedish user (with the same problem).

KentonWhite:
Thanks for the thumbs up.  I all for moving to read.csv2 in the future.  I'll start by adding a warning for now.

PR #196, opened 2017-05-03
Change csv2.reader() to use R's read.csv2() function as announced in #195 and discussed in #170

crubb:
With reference to #195 and #170.

I arbitrarily chose June 2017 for the switch, what do you think? ;)

Best regards,
Christian


KentonWhite:
Let me get this current version of ProjectTemplate on CRAN (with the warning).  We'll then bump the github version and add this change.

crubb:
Good idea.. I suppose giving a version number in the notification messages makes a lot of sense as well ;)

KentonWhite:
In the process of preparing the must recent CRAN release, I noticed I never merged this in!  Sorry, finally merged 😀

Issue #204, opened 2017-08-28
Save and retrieve customized global.dcf 


rsangole:
I would like to request a feature to save a user-customized version of global.dcf someplace. For each project I create, I end up editing the libraries portion of the file and add all my go-to libraries. It's a pain to do everytime, so today, I first open an old project and copy-paste global.dcf into my new project. A bit cumbersome, and some thing we can program our way out of.

Hugovdberg:
There is work underway to create custom templates, which includes the
`global.dcf`, perhaps @connectedblue can give an update on his work.

Op 28 aug. 2017 20:00 schreef "Rahul S" <notifications@github.com>:

connectedblue:
Hi all,

Yes, this use case and others would be easily accommodated with the work I have done on custom template. In fact, that's exactly the reason why I wanted to build this feature.

My progress so far is mixed.  I submitted an initial version which @kentonwhite pointed out would not pass CRAN checks because it wrote files to the R lib directory. We briefly discussed whether it would be possible to create this in a seperate package and link in somehow.

I have such a package in my github which I've been using successfully on projects in my workplace. It has been redesigned to write a config file to the users home directory, which is still in breach of CRAN guidelines. There are also some implementation issues which use non exported functions from devtools which would cause the Travis test to complain.

But the biggest challenge is how to link it back to the master version. I've been very busy in a new start up this year and haven't had much time to progress.

Perhaps I should raise a seperate issue to start a community discussion on this feature.

rsangole:
@connectedblue Thanks for the update. Perhaps one solution is as follows: we develop and maintain a curated list of packages, perhaps a few lists each with a different purpose. These lists could be developed by user surveys or by analyzing existing projects on github. The lists could be selectable in the `create.project()` as well as via a new function, say `load.libraries()`.

For example: 

1. For projects focusing on exploratory data analysis, for the most part, we need: tidyverse, reshape2, lattice, forcats, lubridate, magrittr, corrplot, zoo
2. If we need to perform some modeling, we can add packages like: caret, xgboost, MASS, e1071, pROC etc
3. For projects dealing with quality control, we can add packages: qcc, fitdist... etc

Perhaps the base load.project() comes with (1). Then, if the user needs (2), we can call something like `load.libraries(append = 2).

This gives us flexibility in staying within 1 package, while offering a few options to the users,

Thoughts?

KentonWhite:
[...]

Issue #36, opened 2014-02-05
Replace `minimal` with `template.name` in create.project()


krlmlr:
The parameter `minimal` should be replaced by a parameter `template.name = c("full", "minimal")` to allow adding other templates.

Website documentation has to be updated.

Hugovdberg:
This issue can be closed since #206 was merged. It is implemented as `template = 'full'`, with the option of creating your own templates.

PR #206, opened 2017-08-30
Add custom templates

Hugovdberg:
This is a first try at creating custom templates, as proposed in #204 and, originally, in #36. It seems like a very large change, but in fact the changes can be grouped in a few categories:

- `create.template.R` was added as a lightweight wrapper around `create.project` with some checking for a sensible location
- A global option `ProjectTemplate.templatedir` was introduced to allow the user to specify a directory which contains her or his templates.
- `migrate.template` was added as a wrapper around `migrate.project` to migrate the configuration of the template.
- `create.project` was changed to take an argument `template` instead of `minimal`, which defaults to `full` to preserve the original behavior.
- Most tests had to be updated for the new signature of `create.project`. I removed all references to `minimal = TRUE|FALSE` because just a few tests actually depended on the template used in the test.
- The `inst/default/full` directory was moved to `inst/default/templates/full`, and a new minimal template was created at `inst/default/templates/minimal` to cope with the new template system.

Some stuff has yet to be added, such as documentation for the functions and for the website, but I wanted to show this already to get some feedback before doing so.


KentonWhite:
@Hugovdberg There is a conflict in the `inst/defaults/full/config/global.dcf`.  Can you resolve the conflict please?

Hugovdberg:
I was looking into fixing the problem, but I see that the `inst/defaults/config/default.dcf` still has the old version number, and some other default settings than `inst/defaults/templates/full/config/global.dcf`. Shouldn't those two be the same? And if they should be the same, than how do you feel about removing the `global.dcf` from the base templates provided, and add some logic to `create.project` to copy the `default.dcf` into the project if it's missing in the template? That way we don't have to manage the default settings in both the `full` and `minimal` template.

KentonWhite:
Good point about not having to update the `global.dcf` file in multiple places.  I'm not sure how much effort it would be.  Lets make resolve the conflicts on this PR and then open a new PR to copy to use a single `global.dcf` file.

Hugovdberg:
Apparently `default.dcf` contains some other defaults than `global.dcf`, which the tests do rely on. I reverted those changes, though I'm not exactly sure why those settings were hardcoded in the tests and why the tests assume different settings than the normal defaults.

Hugovdberg:
@KentonWhite @connectedblue did you get a chance to review this already? Is there anything holding us back from implementing this feature?

KentonWhite:
Sorry thought I had merged this.  Thanks for the reminder.

Issue #205, opened 2017-08-28
Replace dplyr, ggplot2, stringr with tidyverse


rsangole:
What are your thoughts in replacing these individual packages with tidyverse? Today, tidyverse will automatically import the core tidyverse packages:
    ggplot2, for data visualisation.
    dplyr, for data manipulation.
    tidyr, for data tidying.
    readr, for data import.
    purrr, for functional programming.
    tibble, for tibbles, a modern re-imagining of data frames.

Anyone working in the 'dplyr' / 'Hadley' framework pretty much requires all of these packages for every project.

KentonWhite:
I happy with that idea.  It should be simply a matter of changing the default config file in the `inst` directory.  Would this be something you could make a pull request for?

rsangole:
Yep, simple enough. I'll get it done today.

rsangole:
@KentonWhite I'd also like to propose shifting from reshape to reshape2 given the performance increases in Hadley's package.

KentonWhite:
That is a good idea.

PR #207, opened 2017-08-30
Issue #205 - Use tidy verse

rsangole:
Changes full/global.dcf. Replaces packages with tidyverse. Reshape with reshape2.

### DIFF
--- a/inst/defaults/full/config/global.dcf
+++ b/inst/defaults/full/config/global.dcf
@@ -8,7 +8,7 @@ munging: TRUE
 logging: FALSE
 logging_level: INFO
 load_libraries: FALSE
-libraries: reshape2, tidyverse, plyr, lubridate
+libraries: reshape2, plyr, tidyverse, stringr, lubridate
 as_factors: TRUE
 data_tables: FALSE
 attach_internal_libraries: FALSE
--- a/inst/defaults/full/config/global.dcf
+++ b/inst/defaults/full/config/global.dcf
@@ -8,7 +8,7 @@ munging: TRUE
 logging: FALSE
 logging_level: INFO
 load_libraries: FALSE
-libraries: reshape, plyr, dplyr, ggplot2, stringr, lubridate
+libraries: reshape2, tidyverse, plyr, lubridate
 as_factors: TRUE
 data_tables: FALSE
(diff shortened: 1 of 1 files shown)


inst/defaults/full/config/global.dcf
```diff
@@ -8,7 +8,7 @@ munging: TRUE
 logging: FALSE
 logging_level: INFO
 load_libraries: FALSE
-libraries: reshape, plyr, dplyr, ggplot2, stringr, lubridate
+libraries: reshape2, tidyverse, plyr, lubridate
```
Hugovdberg: It is not recommended to load `plyr` after `dplyr` and raises a warning, please swap the two arround. Also, `stringr` is installed as a dependency for `tidyverse` but is not actually loaded automatically, so should not be removed.

rsangole: Thanks for the feedback. Making these changes.

rsangole: I'm also tempted to insert `forcats` in there. Thoughts? It's been extremely useful in re-coding factors, grouping factors etc.

Hugovdberg:
The new commit looks reasonable. We do need to think about how we communicate we change to `reshape2`, since it isn't a 100% drop in replacement for `reshape`. `migrate.project` doesn't touch the configuration so that's no issue, but perhaps we can notify users via some other way.
Also, @KentonWhite should the suggested packages in `DESCRIPTION` be updated as well, or are those only for packages used internally in `ProjectTemplate`?

rsangole:
That's a good point. How is this type of communication typically handled for a package update?

KentonWhite:
I convinced myself we don't need to worry.  Legacy projects will be written to use the `reshape` package. In those projects they will have a `global.dcf` file with the `reshape` package not the `reshape2` package.  Only new projects will be created with the `reshape2` package.  By definition they do not have existing code and so nothing to break!  If they want to use `reshape` with a new project they can manually change the config file.

I think it is fine to make a note of the change in the `NEWS.md` file.

rsangole:
Perfect. I've put in reshape2 in the pull request.

Thoughts on also including `forcats`? It's something I use everyday, but not sure about you guys.

Issue #215, opened 2017-12-13
Add Tibble configuration


pomalley08:
Is it possible to add an option to the configuration file to read in data frames as tibbles? I'm thinking this would be similar to the existing option to read in as a data.table. Or would this be best handled in a munging script?

KentonWhite:
I'm not familiar tibbles, but it sounds like it should be possible.  Would you be able to try adding the feature as a pull request?

pomalley08:
Sure, I'll take a look at it.

Hugovdberg:
A tibble is essentially a `data.frame` with a nice print function associated with it. The print function makes sure only as much data is shown as actually fits within the console, and prints helpful info about what's left out.
I think this is a great addition! Should we perhaps not even make it optional, just apply `tbl_df` on all variables? A `data.table` really behaves differently from a normal `data.frame`, a `tibble` only looks different, and I don't think it negatively affects performance.

rsangole:
I think this is a great idea. Tibbles do have other advantages vs conventional dataframes. I would refer readers to http://r4ds.had.co.nz/tibbles.html

pomalley08:
I looked into adding this feature, but it's a bit beyond my ability at this point. Someone else can try if they are interested.

rsangole:
@Hugovdberg Yes. I think making `tibble` the default (and perhaps the only option) is not a bad idea, given where R is headed in terms of using the `tidyverse`. Having said that, we can keep existing code under an option `return_data_frame` to keep backward compatibility for existing code.

Hugovdberg:
@rsangole I don't think we really need to keep an option if it doesn't conflict with `data_tables` (although we could simply don't apply `as_tibble` if `data_tables` is `TRUE`). I'd rather add it to version 0.9 and list the change in `migrate_project`.
@pomalley08 I could add the functionality, but if you'd like I can give you a few pointers first if you would like to learn how to get involved?

rsangole:
Ok makes sense.
@Hugovdberg between you or @pomalley08 , whoever makes the changes, could you put a few notes on where you made the change here? I'd like to learn a bit more on the code structure too.

pomalley08:
@Hugovdberg I appreciate the offer, but unfortunately I don't have the time to commit to this right now.

Hugovdberg:
@rsangole I just remembered that I would describe the changes I made to implement this. There are several links to the lines in `load.project.R` to see which point in the code I'm referring to.

[...]

PR #217, opened 2018-01-23
Feature request #215: Convert data.frame to tibble

Hugovdberg:
As per feature request #215, this PR implements a change to convert all loaded data to `tibble`s. The change is mostly cosmetically as it stands, since it uses a different `print` function from a standard `data.frame`.
For now it doesn't use the full powers of the `tidyverse`, as it simply converts to a `tibble` instead of using the `readr` functions for loading data. The main rationale is that doing so would introduce a major backwards incompatibility as it removes the option `stringsAsFactors` altogether. Given the popularity of the `tidyverse` within the R community, and of course the performance improvements of the `readr` functions over the `base` functions, it is definitely something we as a project should consider. That would be for a version 1.x though I guess.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,6 +48,7 @@ Suggests:
     RPostgreSQL,
     GetoptLong,
     whisker,
+    tibble,
     testthat (>= 0.8)
 URL: http://projecttemplate.net
 BugReports: https://github.com/johnmyleswhite/ProjectTemplate/issues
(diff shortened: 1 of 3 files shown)


Hugovdberg:
Nice, it says my work is bad because Travis CI couldn't even install the R environment, and it seems there is something outdated with the Travis environment. @KentonWhite do you know how to fix this?

PR #232, opened 2018-04-23
Fixes bug with RStudio addins not importing package on click

rsangole:
This commit adds two functions in addins.R which first load the ProjectTemplate package
using require() and then call the appropriate function. Previously, the load.project() and
reload.project() were being referenced directly by the add-ins but this causes an error if
the user has not run require(ProjectTemplate) explicitly before clicking the add-ins buttons.

This commit fixes this bug.

@KentonWhite how do you version control for minor fixes like this? Do you increment X in a.b.X ?


KentonWhite:
I don't worry about versioning for commits on git hub.  I wait until releasing to cran and then bump the version number.  If the release only has non breaking changes I increment the minor version.  If there are breaking changes I increment the major number.

Issue #201, opened 2017-07-07
.RProj creation


rsangole:
I'm requesting a feature in the `create.project()` function that also creates a ProjectName.RProj file (and associated files) to the root folder?

I use ProjectTemplate reguarly now with RStudio, but since the function doesn't create a .RProj itself, it's a two step process for me today - i. Call create.project, and ii. Create a new R project within existing directory.

The advantage of making the work an RStudio project is 
1. Project specific configurations like Code Editing controls, Build tools controls, Git/SVN controls etc
2. RStudio functionality for loading a new project using the dropdown menu on the top right hand side

KentonWhite:
@rsangole This sounds like a very nice feature.  I don't use RStudio personally but do know that many of our users do.  I'm not super familiar with how the .RProj file is supposed to look.  Could you start a pull request with a new folder `inst/defaults/rstudio` that is a copy of `inst/defaults/full` and then place in that folder the .RProj file you would like and where it should be in the directory structure.  We can then modify `create.project()` so that it takes a rstudio flag.

rsangole:
Will do, cheers!

rsangole:
Sent you an email, cheers.

rsangole:
Hello, following up on this request; lost track of time. (But, the feature is still really useful!)

Hugovdberg:
Could this be solved by pr #206? Perhaps it should be amended to rename the
.rproj file if it exists in the template. Rstudio will always show the
directory name as the project name so it isn't technically necessary to
rename the file.

Op 27 nov. 2017 05:53 schreef "Rahul S" <notifications@github.com>:

rsangole:
Yea, I don't see why not. We can create a default template which hosts a basic .RProj file, and write a small function to rename it to the project name. Fits well within #206 framework.

Hugovdberg:
I would suggest not to add another builtin template but either leave it up to the user to create his or her own custom template if an .Rproj file is necessary, or always add the .Rproj to the `full` template. The main advantage of putting it in the custom template is that we avoid discussions on whether we want 2, 4 or 8 spaces by default, want to add a new line at the end of the file if it's missing, remove spaces or not..

rsangole:
That's fair, and it solves the problem at hand.

PR #234, opened 2018-04-24
Closes 201: Adds rstudio.project as a feature to create.project function

rsangole:
Allows the user to initialize the ProjectTemplate directory as an RStudio project as well by creating a projectname.Rproj file with standard defaults in the working directory.

Reference issue #201 .

### DIFF
--- a/R/create.project.R
+++ b/R/create.project.R
@@ -22,6 +22,10 @@
 #'   If \code{"force.empty"}, the target directory must be empty;
 #'   if \code{"allow.non.conflict"}, the method succeeds if no files or
 #'   directories with the same name exist in the target directory.
+#' @param rstudio.project A boolean value indicating whether the project should
+#'   also be an 'RStudio Project'. Defaults to \code{FALSE}. If \code{TRUE},
+#'   then a `projectname.Rproj` with usable defaults is added to the ProjectTemplate
+#'   directory.
 #'
 #' @return No value is returned; this function is called for its side effects.
 #'
@@ -38,7 +42,8 @@
 #'
 #' \dontrun{create.project('MyProject')}
 create.project <- function(project.name = 'new-project', template = 'full',
-                           dump = FALSE, merge.strategy = c("require.empty", "allow.non.conflict"))
+                           dump = FALSE, merge.strategy = c("require.empty", "allow.non.conflict"),
+                           rstudio.project = FALSE)
 {
 
   .stopifproject(c("Cannot create a new project inside an existing one",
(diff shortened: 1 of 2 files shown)


KentonWhite:
Sorry didn't see we were missing tests.  @rsangole can you add some tests please.  

Remind me what the tradeoffs were between having the option and installing your own `.RProj` file.  I saw this as a convenience for RStudio users but as always could be missing something.

rsangole:
Hi @KentonWhite , I'll add some tests, sure.

Installing a `.RProj` isn't viable since each time a project is created, the `.RProj` file would need to be manually renamed to the name of the project. If kept at a default name only (say `RStudioProject.Rproj`, then everything in a dropdown menu like this:
image
would say the same name. Not a usable solution.

There are two manual workarounds right now -

- First create a ProjectTemplate project and then create an RStudio project  using "File > New Project > In existing directory", but then the two checkbox options available in "File > New Project > New Directory" don't show up, which are widely used
image
- OR, first create a project in RStudio, and then create.project with the `allow.non.conflict` option.

Over the last year of adoption of ProjectTemplate and teaching others how to use it in my company, these steps cause the most confusion and reluctance towards the adoption of the package.

KentonWhite:
That is a lot of steps!  @rsangole this PR eliminates those steps?

rsangole:
Yep! I get's very tiring to do too @KentonWhite , I basically use ProjectTemplate, git and RStudio everyday - it's a 100% of my job. And this PR will make life a lot easier for someone like me. Essentially, now the command will be:
`create.project('project_name', rstudio.project = TRUE)`
to make a new ProjectTemplate project + an RStudio enabled project.

rsangole:
@Hugovdberg , sorry missed your previous ping there. I am missing tests - let me think of what tests are needed and add them this weekend.

We did go back and forth on if this is worth it, and based on the past year's experience with both the tools hand in hand, I could justify it. (Look at my post above.)

Hugovdberg:
@rsangole you actually don't need to rename the file, Rstudio always shows the name of the directory not of the `.Rproj` file. I simply have a file `standard.rproj` in my custom template and I don't see any reference to `standard` in the interface. That's why I previously said you don't need any special options. Sorry I didn't make that clear enough.

rsangole:
@Hugovdberg ah, I see what you are saying now. Are you still on the page of the custom template vs in built feature? I still see the in built as a quicker option allowing for easier adoption of the package, so I'd advocate we support this feature.

KentonWhite:
Given the discussion, it is likely that others may also have difficulty with the custom template approach.  Since the work is done for the built-in feature, lets make an exception for RStudio.  Other frameworks should use the custom template solution.

Issue #237, opened 2018-04-28
Feature request: Add feather reader


Hugovdberg:
#### Report an Issue / Request a Feature

I'm submitting a (Check one with "x") : 
- feature request

***

##### Issue Severity Classification - 

(Check one with "x") : 
- 3 - Low

##### Expected Behavior

`.feather` files in the `data/` directory are loaded in memory.

##### Current Behavior

`.feather` files are ignored.

##### Possible Solution

```r
feather.reader <- function(data.file, filename, variable.name) {
    assign(variable.name, feather::read_feather(filename), envir = .TargetEnv)
}
```

rsangole:
@Hugovdberg , any reason why we don't want to read feather files from `cache/`, or is it that conceptually, the "reader" functionality is being defined for the `data/` folder?

KentonWhite:
This is just setting on the `feather` reader.  The cache feature is still being developed!

rsangole:
Got it, thanks!

Issue #225, opened 2018-03-14
Feature Request - Feather format for cache files


rsangole:
#### Report an Issue / Request a Feature

I'm submitting a (Check one with "x") : 
- [ ] bug report 
- [x] feature request

***

##### Issue Severity Classification - 

(Check one with "x") : 
- [ ] 1 - Severe
- [ ] 2 - Moderate
- [x] 3 - Low

##### Expected Behavior

I recommend adding the option of using the `feather` file format as an option when we cache dataframe objects. Read about `feather` [here](https://blog.rstudio.com/2016/03/29/feather/) and [here](https://github.com/wesm/feather). 

Advantages of feather for dataframe objects:

* BLAZING fast compared to RData in R or pickles in Python. I've been using `feather` at work, and it's been really really fast.
* Cached files can be directly accessed from Python (or vice versa) given `feather` was developed by Wes McKinney & Hadley Wickam together. This helps when collaborating on projects with Python aspects.

##### Possible Solution

Potential call:

`cache('dataframe_object', feather = TRUE)`

should save an object called `dataframe_object.feather`.

Hugovdberg:
Since the cache files shouldn't be accessed by users on their own I would rather stick to the convention over configuration philosophy and just introduce it as the default if it provides all functionality we need. It sounds pretty awesome, and like something I could really use in my day to day work 👍 
Also, we really need a reader for `feather` files in that case.

rsangole:
@Hugovdberg the `feather` package has simple functions `write_feather` and `read_feather` which work great.

I don't disagree with the default options of `cache` (since `feather` can only deal with dataframes anyways, it cannot deal with other R objects), but a choice for users of the package who have to deal with large datasets would be extremely useful. Thus I'm proposing that `cache('name', feather = FALSE)` would be the default.

Hugovdberg:
Better performance shouldn't be an option, it should just be implemented right? ;-) Some checks for object types should be built in anyway to make this possible, so I would suggest we add a `tryCatch` to use `write_feather` first, and if it fails we can fallback to `base::save`. The cached objects are stored in a `data.frame` anyway so perhaps we can actually cache everything using `feather`.

I can try to implement this sometime soon, probably over the weekend.

rsangole:
Always want better performance :+1: 

If `cache` always stores `data.frame`, then `feather` is an excellent alternative. 

[...]

PR #239, opened 2018-04-28
Added feather reader

Hugovdberg:
Although implementing `feather` for the cache takes a little more time (#225), adding a reader for the format is quite straightforward (#237). This makes it easier to import output from a python or R script in the feather format.

 - [x] Add `feather.reader`
 - [x] Add test for feather format


rsangole:
Great! Thanks @Hugovdberg . This will be very useful indeed. I like how you use the checklist buttons to make a list and cross things off as you work through the feature. I'm going to adopt this method.

Hugovdberg:
@rsangole Perhaps we can create a pull request template as for new issues so we can remind people to which steps should be taken before a pull request can be merged. Just a simple list of "add functionality",  "add tests", "add code documentation", "add website documentation".

rsangole:
Yep, good idea. Done. Check it out.

Issue #245, opened 2018-04-30
Internal - Complete COC & License


rsangole:
#### Report an Issue / Request a Feature

I'm submitting a (Check one with "x") : 
- [ ] bug report 
- [x] feature request

***

##### Issue Severity Classification - 

(Check one with "x") : 
- [ ] 1 - Severe
- [ ] 2 - Moderate
- [x] 3 - Low

In the interest of completeness, I noticed that the project is missing two items listed <[here](https://github.com/KentonWhite/ProjectTemplate/community)>.

@KentonWhite I didn't want to select the license since you own the project, but perhaps an MIT license makes sense for this project? As for the code of conduct, the default template seems quite good too.

Thoughts?

KentonWhite:
I would follow the tidyverse lead.  What license are they using?

rsangole:
tidyverse uses GNU General Public License

KentonWhite:
I'm fine with the Tidyverse code of conduct.  Thanks @rsangole for asking!

PR #248, opened 2018-05-01
Create LICENSE

rsangole:
#### Types of change

- [ ] New feature (non-breaking change which adds functionality)
- [X] Other (documentation etc)

#### Pull request checklist

 - [X] Add functionality
 - [ ] Add tests

***

#### Proposed changes

- Adds GNU GPL license as discussed in #245


Hugovdberg:
I just realised that `R CMD check` checks the `License` key in the `DESCRIPTION` file. According to the `DESCRIPTION` the project is released under the Artistic 2.0 license. So either we should switch entirely to GPL, or keep the Artistic license. I think we should investigate the implications of the GPL license, I know it is rather strict on the licenses of dependencies and on the usage (it is meant to be a viral license forcing other code to be open as well).

rsangole:
Ah, good catch @Hugovdberg . Will do @KentonWhite . You'll see something shortly.

PR #261, opened 2018-05-05
Change reader signature

Hugovdberg:
Major change of the interface, this is a pull request against the `release-1.0.0` branch so we can clean up the current 0.8.2 release and finalise it into the 0.8.3 release. This feature needs some thorough testing and should not be released in 0.8.3 in my opinion.

#### Types of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Pull request checklist

 - [x] Add functionality
 - [x] Add tests

***

#### Proposed changes

- I created a new version of `.add.extension` that stores the passed in reader with the matching extension. Unlike previous versions an actual reference to the function is stored instead of a character string that is evaluated when loading the data. To be able to store and display the reader nicely in a data.frame I created a new `R6` class that stores the function and the name. 
I then changed `list.data` to actually load it into a tibble because tibbles handle non standard columns better. Also `load.project` actually calls the `reader` function inside the `R6` object instead of looking up the function by name.
- I changed the signature of all the reader functions and add them in the same file to the dispatch table by calling `.add.extension`. This way the reader and the code adding it to the list of available readers is more or less self contained. The `preinstalled.readers` list is now a function that can actually be used to inspect which readers are available. Perhaps this function should be renamed `available.readers`.
- Because `R6` and `tibble` are now actual dependencies of the package I added them to the `Imports` section in `DESCRIPTION`
- Because `xls.reader` and `xlsx.reader` are actually the same reader since we migrated to `readxl` I removed the `xlsx.reader` and added both extensions to the `xls.reader`.


KentonWhite:
@Hugovdberg Is this ready for review or still working on it?

Hugovdberg:
I think it is ready for review, I think further optimization and cleaning is possible, but I prefer to do it after you reviewed it. Keep in mind I propose to only introduce this in 1.0.
Also, the dependency on `tibble` and `R6` should be too much of a problem because they are both dependencies of `dplyr`, which will be introduced with the `tidyverse` anyway. `R6` classes have a nice lightweight profile, and act more like classical OOP where data and methods are kept together. I first tried to implement it as an `S3` class but ended up needlessly cluttering the interface with a lot of auxiliary methods that `R6` just doesn't need.

DESCRIPTION
```diff
@@ -25,6 +25,9 @@ LazyLoad: yes
 Roxygen: list(wrap = FALSE)
 Depends:
     R (>= 2.7)
+Imports:
+    R6,
```
KentonWhite: Is R6 necessary for the new reader signature?  If there is a way to do the same thing with the standard S4 objects I would prefer that.

NAMESPACE
```diff
@@ -23,11 +23,16 @@ export(show.project)
 export(stub.tests)
 export(test.project)
 export(translate.dcf)
+export(type_sum.data_reader)
+importFrom(R6,R6Class)
 importFrom(stats,setNames)
+importFrom(tibble,data_frame)
```
KentonWhite: Just checking — all instances of `data.frame` should now be `data_frame`?

DESCRIPTION
Hugovdberg: Like I said `R6` is an indirect dependency of the `tidyverse` through `dplyr`. There is always a way to do it in S4, but it takes more effort, is a lot less concise and probably harder to understand. The definition of this class, including the formatting and printing function takes 15 lines of code when I include the lines with only parentheses. `R6` is really lightweight, and will be installed on many systems already. I really think if we are going to convert to the `tidyverse` internally this is actually more consistent then trying to implement in `base` functions only.

NAMESPACE
Hugovdberg: I changed them in `list.data`, and since we already convert everything that's not a `data.table` almost all cases should be converted now. We can of course check the rest of the code base for instances that should be converted.

PR #286, opened 2018-11-15
Typo fix within configuring.markdown

MattForshaw:
Typo fix within configuring.markdown


Hugovdberg:
Thank you for fixing this!

PR #318, opened 2025-04-11
add tidyCODE parameter to cache()

gisler:
#### Types of change

#### Pull request checklist

 - [x] Add functionality
 - [ ] Add tests

***

#### Proposed changes

Hi Kenton!

I added the parameter `tidyCODE` to the `cache()` function, which is `TRUE` by default, but can be set to `FALSE` in case `formatR::tidy_source()` fails. I experienced problems with `glue::glue()` presumably in combination with SQL queries containing `\"` lately. Escaped double quotes are often necessary to properly quote identifiers in SQL. When `formatR::tidy_source()` fails to handle such a "complex" query, you can end up with `NA` instead of the tidied query, e.g. `result <- sf::st_read(con, query = glue(NA))`.

Since `formatR::tidy_source()` is there to tidy R code and not SQL, I think it is best to just be able to skip tidying completely if necessary.

P.S. More pull requests are in my pipeline, but don't expect anything exciting. ;⁠)

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -24,6 +24,12 @@ NULL
 #' cached.  Requires suggested package formatR 
 #' @param depends A character vector of other global environment objects that the CODE
 #' depends upon. Caching will be forced if those objects have changed since last caching
+#' @param tidyCODE A logical scalar specifying if the CODE shall be tidied with
+#' the help of \code{\link[formatR]{tidy_source}}. As, for example, whitespace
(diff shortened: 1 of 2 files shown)


KentonWhite (approved):

KentonWhite:
Thanks @gisler!   I've merged into Master on GitHub.   I hope I'll have time this weekend to push up to CRAN.

gisler:
Thank you as well, @KentonWhite.

PR #323, opened 2025-05-13
Added project.directory parameter to create.project.R

3styleJam:
#### Types of change

- [ ] New feature (non-breaking change which adds functionality)

#### Pull request checklist

 - [ ] Add functionality
 - [ ] Add tests

***

#### Proposed changes

- 
-

### DIFF
--- a/R/create.project.R
+++ b/R/create.project.R
@@ -10,6 +10,7 @@
 #'
 #' @param project.name A character vector containing the name for this new
 #'   project. Must be a valid directory name for your file system.
+#' @param project.directory A character vector containing the full path of the project directory up to and excluding the project.name.
 #' @param template A character vector containing the name of the template to
 #'   use for this project. By default a \code{full} and \code{minimal} template
 #'   are provided, but custom templates can be created using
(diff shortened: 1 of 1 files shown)


KentonWhite:
This is breaking quite a bit of the functions when no directory is given.  Can you please run that tests and see where it is breaking?  Looking at the logs it is breaking in these places:

- `load.project`
- `clear.cache`
- `cache`
- `reload.project`
- `test.project`
- `stub.tests`
- `project.config`

3styleJam:
> This is breaking quite a bit of the functions when no directory is given. Can you please run that tests and see where it is breaking? Looking at the logs it is breaking in these places:
> 
> * `load.project`
> * `clear.cache`
> * `cache`
> * `reload.project`
> * `test.project`
> * `stub.tests`
> * `project.config`

Sorry. I am not sure exactly what I should be running, I tried `test_local(path = "./tests/testthat")` and this returned many failures with the same error:

`Error in create.project(test_project): object 'project_directory' not found`

Which suggests the project_directory needs to be set, except in the definition of `create.project()` I have given project.directory a default argument of `getwd()`.

3styleJam:
I've made some more progress and now all that's left is 6 warnings. They are all within "test-migration.R" and they all happen at a call:
```r
expect_warning(suppressMessages(load.project()), ...)
```
With the dots being some regular expression:

1. `"migrate.project"` - There doesn't seem to be any warning message produced by any function in the package that contains "migrate.project", so I am not sure why this test exists at test-migration.R:16:13
2. `"missing the following entries: tables_type"` and `"contains the following unused entries: data_tables"` - At test-migration.R:261:5 and test-migration.R:262:5 . I am not sure why this is triggering a warning.

KentonWhite:
Great progress!  I'll wait for the last few errors in `migrate.project` to be fixed and then prep from CRAN.  Thanks for your work.

KentonWhite:
@3styleJam I'm wanting to push a new version to CRAN soon.   Should I wait for the PR to be finished or push without it?

KentonWhite:
Thanks @3styleJam I'm travelling this week will look into merging next week.  The build is failing due to warnings.  I'll see if I can fix up those warnings (not sure yet what is causing the warnings)

gisler:
@3styleJam Sure, I'll take a look at it this weekend.

gisler:
Hi @3styleJam!

I reviewed your changes to "test-cache.R" and they look good to me. :)

Just one remark concerning the new `project.directory` parameter: You added it right after the `project.name` parameter at the second position. This might break existing code in case a user relies on argument positions. That's why I usually add new parameters to the end of the existing parameters list of a function, but before a possible `...`. However, one could also argue that it is bad practice to rely on argument positions. @KentonWhite You have to decide on this. ;)

Besides, @KentonWhite, since there is a mix of different coding styles in the package by now, which kinda forces contributers to change more than necessary ... What do you think of applying the new formatter [Air](https://posit-dev.github.io/air/) to the existing code base?

3styleJam:
> Hi @3styleJam!
> 
> I reviewed your changes to "test-cache.R" and they look good to me. :)
> 
> Just one remark concerning the new `project.directory` parameter: You added it right after the `project.name` parameter at the second position. This might break existing code in case a user relies on argument positions. That's why I usually add new parameters to the end of the existing parameters list of a function, but before a possible `...`. However, one could also argue that it is bad practice to rely on argument positions. @KentonWhite You have to decide on this. ;)
> 
> Besides, @KentonWhite, since there is a mix of different coding styles in the package by now, which kinda forces contributers to change more than necessary ... What do you think of applying the new formatter [Air](https://posit-dev.github.io/air/) to the existing code base?

Thanks @gisler . Good point concerning the parameter position. I placed it second as it seemed appropriate to put next to `project.name`, but since it is not a required argument for the function to execute then I am equally happy it being placed at the last position too (I rely on argument positions when writing code too). Nice to suggest Air to format the code too, if not that then `styler` and `lintr`.

gisler:
Ah, `styler`. Thanks for reminding me about this package.

